PR #12, opened 2021-01-11
Assignment1

MeguminXIII:
Issues #1-3 complete, Program runs with all questions randomized.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -43,10 +43,12 @@ namespace PrincessBrideTrivia.Tests
 
             Question[] listOne = Program.LoadQuestions("Trivia.txt");
             Question[] listTwo = Program.LoadQuestions("Trivia.txt");
+            Question[] listThree = Program.LoadQuestions("Trivia.txt");
+            Question[] listFour = Program.LoadQuestions("Trivia.txt");
 
-            for(i=0; i < listOne.Length; i++)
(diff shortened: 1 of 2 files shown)


Sithale:
Pull Request title includes "Assignment 1" in the title ✔
Pull Request targets Assignment 1 branch ✔
Issue 1: Application no longer crashes ✔
Issue 2: Unit test properly passes ✔
Extra Credit
Issue 3: Feature request implemented ✔
Issue 3: Feature request unit tested ✔

Great work on the extra credit! Not sure if it matters either, but you have a single line of commented code at the base of your AskQuestion method. Again, nice code!

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -27,6 +27,24 @@ public void LoadQuestions_RetrievesQuestionsFromFile()
...
+            Assert.IsNotNull(questions[i]);
+        }
+
+        [TestMethod]
+        public void LoadQuestions_CheckIfQuestionListIsRandomized()
+        {
+            Question[] listOne = Program.LoadQuestions("Trivia.txt");
+            Question[] listTwo = Program.LoadQuestions("Trivia.txt");
+
+            Assert.AreNotEqual(listOne, listTwo);
```
worseTyler, line 45: Even if these two arrays had the same values they would still not be equal. I believe that an array really stores the address of the first element so unless you had listTwo point to listOne then they would be different

breaman:
Nice work overall on the assignment. Not sure what happened with your pull request, though, but the whole file should not show as being "changed". Not docking you for this since I can't see why this issue happened.

PR #16, opened 2021-01-12
Assignment1

MarkCapka:
… and it should be good to go. Hope to polish over the next coup-le of days for the extra credit concept even if i don't qualify


Zophed:
I will review this.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -5,38 +5,93 @@ namespace PrincessBrideTrivia
...
             int numberCorrect = 0;
+
             for (int i = 0; i < questions.Length; i++)
             {
+                int curIndex;
+                         
+                while(true)
+                {
+                    curIndex = randomizeOrder.Next(1, (questions.Length));
+                   //  if(!CheckArray(usedIndices, (curIndex))) break;
```
Zophed, line 30: When this code is commented out, the program fails to load, citing "Unreachable Code" for line 33.  I figured that was just a left over from testing or the like. Ran perfectly once the comments are removed.

Zophed: Discovered that this while loop does not have an exit condition set up correctly. The commented out if statement creates an infinite loop, the code cannot exit. However, with this code enabled, when the program has run it's course and answered all questions contained within the questions array, the check against the CheckArray method will continuously return as true, causing the if statement to never execute the break command. 

This needs to be reworked in order for the percent correct to display correctly at the end of the program.

Zophed (commented):
Assignment1

    Pull Request title includes "Assignment 1" in the title ❌
    Pull Request targets Assignment 1 branch ✔
    Issue 1: Application no longer crashes ✔
    Issue 2: Unit test properly passes ✔

Extra Credit

    Issue 3: Feature request implemented ❌
    Issue 3: Feature request unit tested ❌

PR #17, opened 2021-01-12
Assignment1

coproduct:
Issues 1, 2


breaman:
Nice work on the assignment, got the unit tests passing and the exception fixed so the application will run.

PR #23, opened 2021-01-19
Assignment2

worseTyler:


worseTyler:
I actually believe that in the CreateLogger method you are not supposed to call the ConfigureFileLogger method, if you did call ConfigureFileLogger in CreateLogger then there would never be an instance where CreateLogger should return null from not having ConfigureFileLogger set. Furthermore, if you called ConfigureFileLogger in CreateLogger it would be fairly redundant because ConfigureFileLogger just sets the filepath, even with error checking that would be 2-3 easily understandable lines of code.

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,8 @@
 {
     public abstract class BaseLogger
     {
+        public string? Name { get; set; }
```
mmwoodfo, line 5:
```suggestion
        public string? ClassName { get; set; }
```
For clarity - `Name` is generic.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Logger
+{
+    using System;
+    using System.IO;
+
+    public class FileLogger : BaseLogger
+    {
+        public FileLogger(string? filePath)
+        {
+            this.FilePath = filePath ?? throw new ArgumentNullException($"FilePath was null in the {nameof(FileLogger)}");
```
mmwoodfo, line 10:
```suggestion
            FilePath = filePath ?? throw new ArgumentNullException($"FilePath was null in the {nameof(FileLogger)}");
```
No need to use the `this` keyword. C# is case sensitive.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,20 @@
...
+        }
+
+        public string FilePath { get; private set; }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            File.AppendAllText(this.FilePath, $"{DateTime.Now} {this.Name} {logLevel} : {message} \n");
```
mmwoodfo, line 17:
```suggestion
            File.AppendAllText(FilePath, $"{DateTime.Now} {Name} {logLevel} : {message} \n");
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,33 @@
...
+            {
+                fileLogger = new FileLogger(FilePath)
+                { Name = className };
+            }
+            else
+            {
+                fileLogger = null;
+            }
+
+#pragma warning disable CS8603 // Want to be able to return null
```
mmwoodfo, line 23:
```suggestion

```
Remove this warning and make the method nullable to return null.

Logger/LogFactory.cs
```diff
@@ -1,11 +1,33 @@
...
-    public class LogFactory
+    using System;
+    using System.IO;
+
+    public static class LogFactory
     {
-        public BaseLogger CreateLogger(string className)
+        private static string? FilePath { get; set; }
+
+        public static FileLogger CreateLogger(string className)
```
mmwoodfo, line 10:
```suggestion
        public static FileLogger? CreateLogger(string className)
```
This will fix the warning CS8603 below.

Logger/Logger.csproj
```diff
@@ -1,7 +1,23 @@
...
+	  <Nullable>enable</Nullable>
+	  <LangVersion>9.0</LangVersion>
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Include="..\.editorconfig" Link=".editorconfig" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="StyleCop.Analyzers" Version="1.1.118">
```
mmwoodfo, line 17: These analyzers will clash with the naming & coding conventions followed in this class. Some of my comments on this PR will require these to be disabled in order to avoid warnings in your code.

Logger/LogFactory.cs
```diff
@@ -1,11 +1,33 @@
 ﻿namespace Logger
 {
-    public class LogFactory
+    using System;
+    using System.IO;
+
+    public static class LogFactory
     {
-        public BaseLogger CreateLogger(string className)
+        private static string? FilePath { get; set; }
```
mmwoodfo, line 8:
```suggestion
        private static string? _FilePath;
```
Typically you don't have private auto-properties, just use a variable.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,20 @@
...
+    public class FileLogger : BaseLogger
+    {
+        public FileLogger(string? filePath)
+        {
+            this.FilePath = filePath ?? throw new ArgumentNullException($"FilePath was null in the {nameof(FileLogger)}");
+        }
+
+        public string FilePath { get; private set; }
+
+        public override void Log(LogLevel logLevel, string message)
```
mmwoodfo, line 15: You may want to consider creating the file if it does not already exist to remove any requirement from the user to do so - but your program works fine as is.

One way to do this would be to use a StreamWriter inside a [using ](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement) statement as such:
```C#
using (StreamWriter streamWriter = new StreamWriter(FilePath, true))
{
      streamWriter.WriteLine($"{DateTime.Now} {Name} {logLevel}: {message}");
}
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,33 @@
...
+
+        public static FileLogger CreateLogger(string className)
         {
-            
-            return null;
+            FileLogger? fileLogger;
+            if (File.Exists(FilePath))
```
mmwoodfo, line 13:
```suggestion
            if (File.Exists(FilePath))
```
You may also want to check if FilePath is null & consider programmatically creating the file as mentioned in a previous comment of mine.

Logger/LogFactory.cs
mmwoodfo: Also, this method body can be optionally simplified to:
```C#
            if (File.Exists(FilePath))
            {
                return new FileLogger(FilePath)
                { Name = className };
            }
            return null;
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,71 @@
 ﻿namespace Logger
 {
+    using System;
+
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger baseLogger, string message, params int[] extraParams)
```
mmwoodfo, line 7:
```suggestion
        public static void Error(this BaseLogger baseLogger, string message, params object[] extraParams)
```
You may consider passing an object array here instead.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,71 @@
...
+        {
+            if (baseLogger == null)
+            {
+                throw new System.ArgumentNullException($"{nameof(baseLogger)} is null");
+            }
+
+            string additionalParams = string.Empty;
+            foreach (int param in extraParams)
```
mmwoodfo, line 15:
```suggestion

```
This becomes unnecessary if you change your `extraParams` to an object array.

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,42 @@
...
+            try
+            {
+                LogFactory.ConfigureFileLogger(filePath);
+                string name = "Name";
+
+                // Act
+                FileLogger fileLogger = LogFactory.CreateLogger(name);
+                Assert.IsNotNull(fileLogger);
+            }
+            finally
```
mmwoodfo, line 25: 👍🏻

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,42 @@
...
+            // Arrange
+            string filePath = Path.GetRandomFileName();
+            FileStream fileStream = new FileStream(filePath, FileMode.OpenOrCreate);
+            try
+            {
+                LogFactory.ConfigureFileLogger(filePath);
+                string name = "Name";
+
+                // Act
+                FileLogger fileLogger = LogFactory.CreateLogger(name);
```
mmwoodfo, line 22:
```suggestion
                FileLogger fileLogger = LogFactory.CreateLogger(nameof(LogFactoryTests));
```

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,42 @@
...
+                File.Delete(filePath);
+            }
+        }
+
+        [TestMethod]
+        public void CreateLogger_ReturnNullGivenNoConfiguration()
+        {
+            // LogFactory logFactory = new LogFactory();
+            string name = "Name";
+            FileLogger fileLogger = LogFactory.CreateLogger(name);
```
mmwoodfo, line 37:
```suggestion
            FileLogger fileLogger = LogFactory.CreateLogger(nameof(LogFactoryTests));
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,32 @@
...
+
+            // Act
+            LogFactory.ConfigureFileLogger(filePath);
+            FileLogger fileLogger = LogFactory.CreateLogger(loggerName);
+            fileStream.Close();
+            fileLogger.Log(logLevel, loggerMessage);
+            string lastLine = File.ReadLines(filePath).Last();
+
+            // Assert
+            Assert.AreEqual($"{DateTime.Now} {loggerName} {logLevel} : {loggerMessage} ", lastLine);
```
mmwoodfo, line 29: Use caution here - if this code takes too long to run your comparison will fail because of the DateTime.Now being set a few seconds too late. This is an unlikely scenario - but something to keep in mind.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,32 @@
...
+            // Arrange
+            string filePath = Path.GetFileName("Test.txt");
+            string loggerName = "TestingLogger";
+            string loggerMessage = "Testing Message On The Logger";
+            LogLevel logLevel = LogLevel.Debug;
+            FileStream fileStream = new FileStream(filePath, FileMode.Append);
+
+            // Act
+            LogFactory.ConfigureFileLogger(filePath);
+            FileLogger fileLogger = LogFactory.CreateLogger(loggerName);
```
mmwoodfo, line 23:
```suggestion
            FileLogger fileLogger = new FileLogger(filePath);
```
We are not testing the functionality of the `LogFactory` here, you want to keep your unit tests specific to the components you are testing. In this case we are testing the FileLogger, so it would be better to directly instantiate a new FileLogger rather than relying on the CreateLogger method which could cause unrelated failures.

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -26,23 +29,114 @@ public void Error_WithData_LogsMessage()
...
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Debug_WithNullLogger_ThrowsException()
+        {
+            // Arrange
+            TestLogger? logger = null;
+
+            // Act
+#pragma warning disable CS8604 // Want to make sure that null is caught
```
mmwoodfo, line 108:
```suggestion

```
You don't need to suppress this warning if you mark the `BaseLogger` parameter as nullable in your extension methods.

PR #32, opened 2021-01-25
Assignment3

Zophed:
Almost finished with the assignment, currently everything is running. Looking into how to use Moq, haven't really been able to apply it to this specific situation. Need to finish at least one unit test, but code coverage analysis says 90% of Jester is covered currently.

### DIFF
--- a/CanHazFunny/CanHazFunny/Jester.cs
+++ b/CanHazFunny/CanHazFunny/Jester.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny
+{
+    public class Jester:IJokeService,IJokeOutput
+    {
+        private JokeOutputService outputService;
+        private JokeService jokeService;
+        public JokeOutputService OutputService { get => outputService; set => outputService = value; }
+        public JokeService JokeService { get => jokeService; set => jokeService = value; }
+
+        public Jester(JokeOutputService outputService, JokeService jokeService)
+        {
+            if (outputService != null && jokeService != null)
+            {
+                this.OutputService = outputService;
+                this.JokeService = jokeService;
+            }
+            else
+            {
+                throw new ArgumentNullException(nameof(outputService));
+            }
+        }
+
+        public void TellJoke()
+        {
+            string joke = JokeService.GetJoke();
+            while(joke.Contains("Chuck") || joke.Contains("Norris"))
+                joke = JokeService.GetJoke();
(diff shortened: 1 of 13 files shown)


alexvthornton:
I will review this

alexvthornton:
# Assignment

- `Jester` class takes in the correct interfaces: ✔
- `Jester` dependencies are null checked: ✔
- `Jester` appropriately retrieves jokes and sends them to the output: ✔
- Chuck Norris jokes get filtered out: ✔
- Jester code coverage is above 90%: ✔

### Extra Credit
- Output to console implementation is unit tested: ❌
- `JokeService` uses JSON: ❌

API sends joke in JSON formate but you don't deconstruct the json to only return the joke, instead, you returned the whole thing

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny
+{
+    public class Jester
+    {
+        private IJokeOutput outputService;
+        private IJokeService jokeService;
```
Keboo, line 12: Field should be prefixed with an underscore and PascalCase

```suggestion
        private IJokeOutput _OutputService;
        private IJokeService _JokeService;
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,37 @@
...
+        public IJokeOutput JokeOutputService { get => outputService; set => outputService = value; }
+        public IJokeService JokeService { get => jokeService; set => jokeService = value; }
```
Keboo, line 14: Given that these properties don't add any logic in the getter or setter a simply auto property would be cleaner.
```suggestion
        public IJokeOutput JokeOutputService { get; set; }
        public IJokeService JokeService { get; set; }
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,37 @@
...
+        public Jester(IJokeOutput outputService, IJokeService jokeService)
+        {
+            if (outputService != null && jokeService != null)
+            {
+                this.JokeOutputService = outputService;
+                this.JokeService = jokeService;
+            }
+            else
+            {
+                throw new ArgumentNullException(nameof(outputService));
```
Keboo, line 25: This is a bit of a misleading exception. If the `outputService` is specified and the `jokeService` is null the exception claims that `outputService` is the problem. Typically in the unit test, you would also assert on the `ParamName` or the `ArgumentNullException` to make sure you catch these.

PR #34, opened 2021-01-26
Assignment3

MeguminXIII:


CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny
+{
+    public class Jester
+    {
+        private IJokeOutput jokeOutput;
+        private IJokeService jokeService;
```
Keboo, line 12: Fields should be named with an underscore prefix and PascalCase
```suggestion
        private IJokeOutput _JokeOutput;
        private IJokeService _JokeService;

```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public IJokeOutput JokeOutput { 
+            get => jokeOutput; 
+            private set => jokeOutput = value ?? throw new ArgumentNullException(jokeOutput + "is null");
```
Keboo, line 16: The `ArgumentNullException` constructor that takes in a string, expect the string to the the parameter name, not a message.
```suggestion
            private set => jokeOutput = value ?? throw new ArgumentNullException(nameof(value));
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public IJokeService JokeService
+        {
+            get => jokeService;
+            private set => jokeService = value ?? throw new ArgumentNullException(jokeService + "is null");
```
Keboo, line 22:
```suggestion
            private set => jokeService = value ?? throw new ArgumentNullException(nameof(value));

```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public Jester(IJokeOutput jokeOutput, IJokeService jokeService)
+        {
+            if(jokeOutput == null || jokeService == null)
+            {
+                throw new ArgumentNullException(jokeOutput + " or " + jokeService + " is null");
```
Keboo, line 29: In general you want a separate ArgumentNullException for each case, rather than putting it into one. If you put your cursor on the parameter name, and hit Ctrl+. (lightbulb or auto code button) it will generate these for you.

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -2,13 +2,13 @@
...
 {
-    public class JokeService
+    public class JokeService : IJokeService
     {
         private HttpClient HttpClient { get; } = new();
 
         public string GetJoke()
         {
-            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
+            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api?format=json").Result;
```
Keboo, line 11: This will certainly get the json string back, but it doesn't actually parse the json back into just a joke string.

test.txt
```diff
@@ -0,0 +1 @@
+This is a test file
```
Keboo, line 1: Did you mean to include this?

PR #36, opened 2021-01-26
Assignment3

Sithale:
Main homework goals should be satisfied.
Geek Jokes API should be using the JSON format.

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JesterTests.cs
@@ -1,10 +1,97 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using Moq;
 
 namespace CanHazFunny.Tests
 {
     [TestClass]
     public class JesterTests
     {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Jester_NullJokeOutput_ArgumentNullException()
+        {
+            // Assign
+            _ = new Jester(new JokeService(), null);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Jester_NullJokeService_ArgumentNullException()
+        {
+            // Assign
+            _ = new Jester(null, new JokeOutput());
+        }
+
+        [TestMethod]
+        public void Jester_WorkingJokeOutput()
+        {
+            // Assign
+            IJokeService jokeServ = new JokeService();
+            IJokeOutput jokeOut = new JokeOutput();
+
+            Jester jester = new Jester(jokeServ, jokeOut);
+
+            // Act
+
+            // Assert
+            Assert.AreEqual(jokeOut, jester.JokeOut);
+        }
+
+        [TestMethod]
+        public void Jester_WorkingJokeService()
+        {
+            // Assign
+            IJokeService jokeServ = new JokeService();
+            IJokeOutput jokeOut = new JokeOutput();
+
+            Jester jester = new Jester(jokeServ, jokeOut);
+
+            // Act
+
+            // Assert
+            Assert.AreEqual(jokeServ, jester.JokeServ);
+        }
+
+        [TestMethod]
+        public void TellJoke_ChuckNorrisTest_FindNewJoke()
+        {
+            // Assign
+            Mock<IJokeService> mock = new Mock<IJokeService>();
+
+            mock.SetupSequence(jokeService => jokeService.GetJoke())
+                .Returns("Chuck Norris does not sleep. He waits.")
+                .Returns("Chuck Norris can divide by zero, and counted to infinity 3 times.")
+                .Returns("When Chuck Norris falls in water, Chuck Norris doesn't get wet. Water gets Chuck Norris.")
+                .Returns("No joke can stand up to this competition!")
+                .Returns("Just any other old knee-slapper.");
+
+            Jester jester = new Jester(mock.Object, new JokeOutput());
+
+            // Act
+            jester.TellJoke();
+
+            // Assert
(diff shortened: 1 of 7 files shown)


MeguminXIII:
I will review

MeguminXIII (commented):
Jester class takes in the correct interfaces: ✔
Jester dependencies are null checked: ✔
Jester appropriately retrieves jokes and sends them to the output: ✔
Chuck Norris jokes get filtered out: ✔
Jester code coverage is above 90%: ❌
Says 77.78%. 		set_JokeOutput(CanHazFunny.IJokeOutput)	6	75.00%	2	25.00%
                		set_JokeService(CanHazFunny.IJokeService)	6	75.00%	2	25.00%
These are the blocks that are not covered. they have 25% coverage and 75% not covered. which brings the jester class down to 77.78%
Extra Credit
Output to console implementation is unit tested: ❌
has a JokeOutputToConsole test, but I think that only tests that the jester.TellJoke has an output. I could be wrong however.
JokeService uses JSON: ✔

good job

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,97 @@
...
+            mock.Verify(jokeService => jokeService.GetJoke(), Times.Exactly(4));
+        }
         
+        [TestMethod]
+        public void PrintJoke_JokeOutputToTheConsole_JokeIsPrinted()
+        {
+            //Assign
+            Mock<IJokeService> mockService = new Mock<IJokeService>();
+            mockService.SetupSequence(JokeService => JokeService.GetJoke())
+                .Returns("Just any other old knee-slapper.");
```
Keboo, line 84: If you only need to return a single value you can use `Setup` rather than `SetupSequence`
```suggestion
            mockService.Setup(JokeService => JokeService.GetJoke())
                .Returns("Just any other old knee-slapper.");
```

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,97 @@
...
+
+            Mock<IJokeOutput> mockOutput = new Mock<IJokeOutput>();
+            mockOutput.SetupSequence(JokeOutput => JokeOutput.TellJoke("Just any other old knee-slapper."));
```
Keboo, line 87: We will cover this later in the class when we get to lambdas but we typically name the parameter with camelCase
```suggestion
            mockOutput.SetupSequence(jokeOutput => jokeOutput.TellJoke("Just any other old knee-slapper."));
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+
+namespace CanHazFunny
+{
+    public class Jester
+    {
+        private IJokeService JokeService;
+        private IJokeOutput JokeOutServ;
```
Keboo, line 8: Field should be named with underscore PascalCase
```suggestion
        private IJokeService _JokeService;
        private IJokeOutput _JokeOutServ;
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public IJokeOutput JokeOut { get => JokeOutServ; set => JokeOutServ = value; }
+        public IJokeService JokeServ { get => JokeService; set => JokeService = value; }
```
Keboo, line 11: Because these properties don't add any logic in their getters or setters they could simply be auto properties and remove the need for fields.
```suggestion
        public IJokeOutput JokeOut { get; set; }
        public IJokeService JokeServ { get; set; }
```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public Jester(IJokeService jokeService, IJokeOutput outJoke)
+        {
+            if(jokeService == null || outJoke == null)
+            {
+                throw new ArgumentNullException(nameof(jokeService));
```
Keboo, line 17: This is not quite right. If `jokeService` is specified and `outJoke` is null the thrown exception indicates the wrong argument name (since it always sets it to `jokeService`). When checking parameters for null, you want to check each individually.

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -2,13 +2,14 @@
...
-    public class JokeService
+    public class JokeService : IJokeService
     {
         private HttpClient HttpClient { get; } = new();
 
         public string GetJoke()
         {
-            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
+            //string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api").Result;
+            string joke = HttpClient.GetStringAsync("https://geek-jokes.sameerkumar.website/api?format=json").Result;
```
Keboo, line 12: This returns the json string, but it doesn't actually parse the json into the normal joke string.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,38 @@
...
+            }
+            else
+            {
+                this.JokeOut = outJoke;
+                this.JokeService = jokeService;
```
Keboo, line 22: This is actually assigning the field not the property. This is why we favor just using auto properties rather than always having backing fields for them.
```suggestion
                this.JokeServ = jokeService;
```

Keboo (commented):
I am not seeing a code review from you for assignment 3. Am I missing it?

Keboo:
When I check the code coverage for the Jester class it shows as 96% :+1:
image

PR #38, opened 2021-01-26
Assignment3

cmartinezaraujo:


coproduct:
• Jester class takes in the correct interfaces: :heavy_check_mark: 
• Jester dependencies are null checked: :heavy_check_mark: 
• Jester appropriately retrieves jokes and sends them to the output: :heavy_check_mark: 
• Chuck Norris jokes get filtered out: :heavy_check_mark: 
• Jester code coverage is above 90%: :heavy_check_mark: 
Extra Credit
• Output to console implementation is unit tested: :x: 
• JokeService uses JSON: :x:

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,10 +1,73 @@
...
+        public void Jester_TakesValidServices()
+        {
+            JokeService jokeService= new();
+            JokeDisplay jokeDisplay = new();
+
+            Jester joker = new Jester(jokeService, jokeDisplay);
+
+            Assert.AreEqual(joker.JokeService,jokeService);
+            Assert.AreEqual(joker.DisplayService, jokeDisplay);
+        }
```
Keboo, line 34: This works, but a simpler solution is to make the properties private so you don't have to unit test them

CanHazFunny/CanHazFunny/IJokeDisplay.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny
+{
+    public interface IJokeDisplay
+    {
+        public void DisplayJoke(string joke);
```
Keboo, line 11: Avoid specifying access modifiers on interfaces
```suggestion
        void DisplayJoke(string joke);

```

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,37 @@
...
+            _JokeService = jokeService ?? throw new System.ArgumentNullException(nameof(jokeService));
+
+            _DisplayService = tellerService ?? throw new System.ArgumentNullException(nameof(tellerService));
+        }
+
+        public void TellJoke()
+        {
+            string joke = JokeService.GetJoke();
+
+            while(joke.Contains("Chuck") | joke.Contains("Norris"))
```
Keboo, line 29: Though this technically works, there is a subtle difference between `|` and `||`. The first is a bitwise or, the second is a logical or. Unless you are looking for bit-wise operations, most of the time you want to use `||`
```suggestion
            while(joke.Contains("Chuck") || joke.Contains("Norris"))

```

PR #42, opened 2021-01-30
Assingment4

Zophed:
Have finished the main portion of assignment. There is a very small portion of the unit tests that weren't completed, ~6.3% of both my written files, the NumSet and Writer classes.  Did not complete the extra credit. Might do it if I have time later this weekend.

### DIFF
--- a/NumSet/NumSet.Tests/NumSetTests.cs
+++ b/NumSet/NumSet.Tests/NumSetTests.cs
@@ -1,0 +1,195 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+
+namespace Assingment4.Tests
+{
+    [TestClass]
+    public class NumSetTests
+    {
+        [TestMethod]
+        public void Constructor_PassesParamsArray_AssignsParamsArrayToHashSet()
+        {
+            //Assign
+            int[] intArray = new int[] { 1, 3, 2 };
+            HashSet<int> temp = new HashSet<int>(intArray);
+
+            //Act
+            NumSet numSet = new NumSet(intArray);
+
+            //Assert
+            Assert.IsNotNull(numSet.IntSet);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Constructor_PassedNullArgs_ThrowsException()
+        {
+            var numSet = new NumSet(null);
+        }
+
+        [TestMethod]
+        public void ReturnArray_TakesLocalHashSet_ReturnsAnIntArray()
+        {
+            //Assign
+            int[] intArray = new int[] { 1, 3, 2 };
+            NumSet numSet = new NumSet(intArray);
+
+            //Act
+            int[] returnArray = numSet.ReturnArray();
+
+            //Assert
+            CollectionAssert.AreEqual(intArray, returnArray);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ReturnArray_HashSetHasNotBeenSetupByConstructor_ThrowsException()
+        {
+            //Assign
+            int[] intArray = new int[] { 1, 3, 2 };
+            NumSet numSet = new NumSet(intArray);
+            numSet.IntSet = null;
+
+            //Act
+            numSet.ReturnArray();
+
+            //Assert
+
+        }
+
+        [TestMethod]
+        public void Equals_PassedTwoDifferentNumSet_ReturnsFalse()
+        {
+            //Assign
+            int[] intArray1 = new int[] { 1, 2, 3 };
+            int[] intArray2 = new int[] { 1, 2, 4 };
+            NumSet numSet1 = new NumSet(intArray1);
+            NumSet numSet2 = new NumSet(intArray2);
+
+            //Act
+
+            //Assert
+            Assert.IsFalse(numSet1.Equals(numSet2));
+        }
+
+        [TestMethod]
+        public void Equals_PassedTheSameNumSet_ReturnsTrue()
+        {
+            //Assign
+            int[] intArray = new int[] { 1, 2, 3 };
+            NumSet numSet = new NumSet(intArray);
+
+            //Act
+
+            //Assert
+            Assert.IsTrue(numSet.Equals(numSet));
+        }
+
+        [TestMethod]
+        public void Equals_PassedNonNumSetObject_ReturnsFalse()
+        {
+            //Assign
+            int[] intArray = new int[] { 1, 2 };
+            NumSet numSet = new NumSet(intArray);
+            Object obj = new object();
+
+            //Act
+
+            //Assert
+            Assert.IsFalse(numSet.Equals(obj));
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Equals_PassedNullIntSet_ThrowsException()
+        {
+            //Assign
+            int[] intArray1 = new int[] { 1, 2, 3 };
+            int[] intArray2 = new int[] { 1, 2, 4 };
+            NumSet numSet1 = new NumSet(intArray1);
+            NumSet numSet2 = new NumSet(intArray2);
+            numSet1.IntSet = null;
+
+            //Act
+            bool temp = numSet1.Equals(numSet2);
+
+            //Assert
+
+        }
+
+        [TestMethod]
+        public void EqualityAndInequality()
+        {
+            //Assign
+            int[] intArray1 = new int[] { 1, 2, 3 };
+            int[] intArray2 = new int[] { 1, 2, 4 };
+            NumSet numSet1 = new NumSet(intArray1);
+            NumSet numSet2 = new NumSet(intArray2);
+            NumSet? numSet3 = null;
+            NumSet? numSet4 = null;
+
+            //Act
+
+            //Assert
+            Assert.IsTrue(numSet3 == numSet4);
+            Assert.IsTrue(numSet1 != numSet2);
+            Assert.IsFalse(null == numSet1);
+            Assert.IsFalse(numSet1 == null);
+        }
+
+        [TestMethod]
+        public void ToString_ReturnsStringOfHashSet()
+        {
+            //Assign
+            int[] intArray = new int[] { 1, 2 };
+            NumSet numSet = new NumSet(intArray);
+
+            //Act
+            string temp = numSet.ToString();
+
+            //Assert
+            Assert.AreEqual(temp, "1, 2");
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ToString_NullHashSet_ThrowsException()
+        {
+            NumSet? numSet = new NumSet();
+            numSet.IntSet = null;
+            string v = numSet.ToString();
+        }
+
+        [TestMethod]
+        public void GetHashCode_ReturnsCorrectHashCode()
+        {
+            //Assign
+            int[] intArray = new int[] { 1, 2 };
+            NumSet numSet1 = new NumSet(intArray);
+            NumSet numSet2 = numSet1;
+
+            //Act
+            int hashCode1 = numSet1.GetHashCode();
+            int hashCode2 = numSet2.GetHashCode();
+
+            //Assert
+            Assert.IsTrue(hashCode1 == hashCode2);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void GetHashCode_NullSet_ThrowsException()
+        {
+            //Assert
+            NumSet numSet = new NumSet();
+            numSet.IntSet = null;
+
+            //Act
+            int temp = numSet.GetHashCode();
+
+            //Assert
+
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


alexvthornton:
# Assignment
- Both projects created with appropriate properties: ✔
- NumSet constructor accepts variable arguments: ✔
- NumSet ToString implemented correctly: ✔
- NumSet equality operators implemented correctly: ✔
- NumSet fully unit tested: ✔
- SeWriter created in nested namespace: ✔
- SeWriter writes set to file: ✔
- SeWriter properly implements IDisposable: ✔
- SeWriter fully unit tested: ✔

## Extra Credit
NumSet contains cast operator that is unit tested: ❌
Projects updated to properly use Directory.Build.props: ❌

I think you're not supposed to throw any exceptions in the dispose method but other than that looks good.

NumSet/NumSet/NumSet.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                return false;
+            }
+
+            if (IntSet is null)
+            {
+                throw new ArgumentNullException(nameof(IntSet), "IntSet is null in NumSet.Equals(object obj)");
+            }
+
+            return this.IntSet == ((NumSet)obj).IntSet;
```
mmwoodfo, line 57: This will invoke the `==` operator which you have defined below with the following code to call the `Equals` method.
```C#
// Code from your == method on line 92
return temp1.Equals(temp2);
```
This will cause a circular dependency.

NumSet/NumSet.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            //Assert
+            Assert.IsFalse(numSet1.Equals(numSet2));
+        }
+
+        [TestMethod]
+        public void Equals_PassedTheSameNumSet_ReturnsTrue()
+        {
+            //Assign
+            int[] intArray = new int[] { 1, 2, 3 };
+            NumSet numSet = new NumSet(intArray);
```
mmwoodfo, line 81: You'll want to set up a second `NumSet` object rather than comparing the same object to its self.
```suggestion
            NumSet numSet1 = new NumSet(intArray);
            NumSet numSet2 = new NumSet(intArray);
```

NumSet/NumSet.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            //Assert
+
+        }
+
+        [TestMethod]
+        public void Equals_PassedTwoDifferentNumSet_ReturnsFalse()
+        {
+            //Assign
+            int[] intArray1 = new int[] { 1, 2, 3 };
+            int[] intArray2 = new int[] { 1, 2, 4 };
```
mmwoodfo, line 66:
```suggestion
            int[] intArray2 = new int[] { 1, 2, 3 };
```
If you changed your second array to match your first array as I've done above this test method will still pass, which is not what you want because now the two NumSets are equal to eachother based on the assignments criteria.

NumSet/NumSet.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            NumSet numSet2 = new NumSet(intArray2);
+
+            //Act
+
+            //Assert
+            Assert.IsFalse(numSet1.Equals(numSet2));
+        }
+
+        [TestMethod]
+        public void Equals_PassedTheSameNumSet_ReturnsTrue()
```
mmwoodfo, line 77:
```suggestion
        [DataTestMethod]
        [DataRow(new int[] { 1, 3, 2 }, new int[] { 1, 2, 3 })] // Should be equal
        [DataRow(new int[] { 3, 2, 1 }, new int[] { 1, 3, 3, 2 })] // Should be equal
        [DataRow(new int[] { 1, 2, 3 }, new int[] { 1, 2, 3 })] // Should be equal
        public void Equals_PassedTheSameNumSet_ReturnsTrue(int[] set1, int[] set2)
```
You may consider making some of these tests `DataTestMethods` so you can test a few different inputs.

NumSet/NumSet.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            NumSet numSet1 = new NumSet(intArray1);
+            NumSet numSet2 = new NumSet(intArray2);
+            NumSet? numSet3 = null;
+            NumSet? numSet4 = null;
+
+            //Act
+
+            //Assert
+            Assert.IsTrue(numSet3 == numSet4);
+            Assert.IsTrue(numSet1 != numSet2);
```
mmwoodfo, line 136: This would still return true if both arrays had the same values - which is not what we want.

NumSet/NumSet.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            int[] intArray2 = new int[] { 1, 2, 4 };
+            NumSet numSet1 = new NumSet(intArray1);
+            NumSet numSet2 = new NumSet(intArray2);
+            NumSet? numSet3 = null;
+            NumSet? numSet4 = null;
+
+            //Act
+
+            //Assert
+            Assert.IsTrue(numSet3 == numSet4);
```
mmwoodfo, line 135: It would be nice to see the `==` operator used on two non-null sets.

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assingment4.Writer
+{
+    public class SetWriter : IDisposable
+    {
```
mmwoodfo, line 11: What you have set up here is a folder, not a nested namespace.

```suggestion
namespace Assingment4
{
    namespace NestedAssignment4 // This is your nested namespace
    {
        public class SetWriter : IDisposable
        {
```

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
...
+        private bool _DisposedValue;
+
+        private string? _filePath;
+        public string? FilePath { get => _filePath; set => _filePath = value; }
```
mmwoodfo, line 15: You don't need this public auto property. You actually don't need to store the file path at all in this class.
```suggestion
```

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        private StreamWriter? fileWriter;
+        public StreamWriter? FileWriter { get => fileWriter; set => fileWriter = value; }
```
mmwoodfo, line 18: You don't need a private variable and a public auto property. You only need one or the other. In this case, you really only want the private variable. Just make sure to name it properly using PascalCase. Also, you don't need to make it nullable since you will be creating it in the constructor.
```suggestion
        private StreamWriter _FileWriter;
```

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        
+        public SetWriter(string? filePath)
+        {
+            FilePath = filePath ?? throw new ArgumentNullException(nameof(filePath));
+            FileWriter = new StreamWriter(FilePath);
```
mmwoodfo, line 24:
```suggestion
            _FileWriter = new StreamWriter(filePath);
```

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        public void WriteToFile(NumSet numSet)
+        {
+            if(numSet == null || numSet.IntSet == null)
+            {
+                throw new ArgumentNullException(nameof(numSet), "SetWriter.WriterToFile() -- null param");
+            }
+
+            if(FileWriter == null)
```
mmwoodfo, line 34: `_FileWriter` shouldn't be null if the above comments are followed.

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                throw new ArgumentNullException(nameof(FileWriter), "SetWriter.WriterToFile() -- StreamWriter not set up");
+            }
+
+            FileWriter.Write(numSet.ToString());
```
mmwoodfo, line 39: You may want to use `WriteLine` here.

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assingment4.Writer
+{
+    public class SetWriter : IDisposable
+    {
+        private bool _DisposedValue;
```
mmwoodfo, line 12: You can set this to be false by default.
```suggestion
        private bool _DisposedValue = false;
```

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
...
+            if (!_DisposedValue)
+            {
+                if (disposing)
+                {
+                    if(FileWriter == null)
+                    {
+                        throw new ArgumentNullException(nameof(FileWriter), "SetWriter.Dispose -- FileWrtier is null");
+                    }
+
+                    FileWriter.Close();
```
mmwoodfo, line 52:
```suggestion
                    FileWriter.Dispose();
```
See the [Microsoft documentation](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose?redirectedfrom=MSDN#implement-the-dispose-pattern) for the dispose pattern.

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
...
+                // TODO: free unmanaged resources (unmanaged objects) and override finalizer
+                // TODO: set large fields to null
+                _DisposedValue = true;
+            }
+        }
+
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+            Dispose(disposing: true);
```
mmwoodfo, line 64:
```suggestion
            Dispose(true);
```

NumSet/NumSet.Tests/Writer/SetWriterTests.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    [TestClass]
+    public class SetWriterTests
+    {
+        [TestMethod]
+        public void Constructor_CreatesNewStreamWriter()
+        {
+            //Assign
+            string temp = Path.GetTempFileName();
+            SetWriter setWriter = new SetWriter(temp);
```
mmwoodfo, line 16:
```suggestion
            using SetWriter setWriter = new SetWriter(temp);
```

NumSet/NumSet.Tests/Writer/SetWriterTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void WriteToFile_PassesNullArgument_ThrowsException()
+        {
+            string temp = Path.GetTempFileName();
+            SetWriter setWriter = new SetWriter(temp);
+            NumSet numSet = new NumSet();
+            numSet.IntSet = null;
+
+            setWriter.WriteToFile(numSet);
```
mmwoodfo, line 57:
```suggestion
            try
            {
                setWriter.WriteToFile(numSet);
            }
            catch
            {
                throw;
            }
            finally
            {
                setWriter.Dispose();
            }
```
Make sure you dispose of.

NumSet/NumSet.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Constructor_PassedNullArgs_ThrowsException()
+        {
+            var numSet = new NumSet(null);
+        }
+
+        [TestMethod]
+        public void ReturnArray_TakesLocalHashSet_ReturnsAnIntArray()
```
mmwoodfo, line 32: It would be nice to have a test that verifies changing the array does not affect the HashSet it was created from.

NumSet/NumSet/Writer/SetWriter.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assingment4.Writer
```
mmwoodfo, line 8: I do see in VisualStudio that you have a folder set up for this nested namespace. You don't actually need to create a folder for it, you can just add `.Writer` to the end of it.

NumSet/NumSet/NumSet.cs
```diff
@@ -0,0 +1,107 @@
...
+                return false;
+            }
+
+            /*
+             * This if statement was used because the compiler was 
+             *  whining that temp1 could be null, this was the fastest
+             *  method of getting it to shut up.
+             */
+            if (temp1 is null)
+                throw new ArgumentNullException(nameof(temp1), "Somehow, a null value got passed the two if statements in the overridden '==' operator. Bring a priest, demonic forces are about.");
```
mmwoodfo, line 91: This can just return false.

PR #46, opened 2021-02-02
Assignment4

MeguminXIII:


worseTyler:
I will review this

worseTyler:
Both projects created with appropriate properties: ✔
NumSet constructor accepts variable arguments: ✔
NumSet ToString implemented correctly: ✔
NumSet equality operators implemented correctly: ✔
NumSet fully unit tested: ❌
SeWriter created in nested namespace: ❌
SeWriter writes set to file: ✔
SeWriter properly implements IDisposable: ✔
SeWriter fully unit tested: ✔
Extra Credit
NumSet contains cast operator that is unit tested: ❌ Projects updated to properly use Directory.Build.props: ❌
When I ran your unit tests some of your NumSet tests failed

Assignment4/Assignment4/SetWriter.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Assignment4
```
worseTyler, line 6: Not nested

Assignment4/Assignment4.Tests/SetWriterTests.cs
```diff
@@ -0,0 +1,60 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assignment4.Tests
```
worseTyler, line 9: This should be nested too I believe

Assignment4/Assignment4.Tests/SetWriterTests.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Assignment4.Tests
+{
+    [TestClass]
+    public class SetWriterTests
+    {
+
+        void setup()
```
worseTyler, line 15: Should uppercase to Setup(), also I'm not entirely sure what this will accomplish in your tests because you never actually use the name "test.txt"

Assignment4/Assignment4.Tests/SetWriterTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            SetWriter setWriter = new SetWriter(path!);
+            setWriter.WriteSetToFile(one);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Constructor_PassingInNull_ThrowsNullException()
+        {
+            SetWriter sw = new(null!);
```
worseTyler, line 38: You can change sw to _

Assignment4/Assignment4/NumSet.cs
```diff
@@ -0,0 +1,65 @@
...
+            int num = 7;
+            foreach (int i in Set!)
+                num += i * 43;
+            return num;
+        }
+
+        public static bool operator ==(NumSet one, NumSet two)
+        {
+            if (one is null && two is null) return true;
+            if (one is null ^ two is null) return false;
```
worseTyler, line 49: Probably worth adding a unit test to test this case here

Assignment4/Assignment4.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        public void ReturnArray_WhenCalled_ReturnsAnArray()
+        {
+            int[] array = new int[] { 1, 2, 3, 12 };
+
+            NumSet arraySet = new NumSet(array);
+
+            Assert.IsNotNull(arraySet.ReturnArray());
+        }
+
+        [TestMethod]
```
mmwoodfo, line 37: You can also use the [[DataRows] attribute](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.datarowattribute?view=mstest-net-1.3.2) here.

Assignment4/Assignment4.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        public void Equals_ReturnsValidValue()
+        {
+            NumSet one = new(1, 2, 3, 5, 6);
+            NumSet two = new(1, 2, 3, 5, 6);
+            NumSet three = new(9, 10, 11, 12);
+
+            Assert.IsTrue(one.Equals(two));
+            Assert.IsFalse(one.Equals(three));
+            Assert.IsTrue(one.Equals(one));
+        }
```
mmwoodfo, line 47: To get this test to pass you can switch from using `==` in your `Equals` method to using `SetEquals`.

Assignment4/Assignment4.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        [TestMethod]
+        public void NotEqualsOperator_ReturnsValidValue()
+        {
+            NumSet one = new(1, 2, 3, 5, 6);
+            NumSet two = new(1, 2, 3, 5, 6);
+            NumSet three = new(9, 10, 11, 12);
+
+            Assert.IsTrue(one != three);
+        }
```
mmwoodfo, line 90: This does pass - but falsely. As we can see from the failed tests, `Equals` will always return `false`. (Unless the two objects are the same, eg: `NumSet three = one` would cause this test to fail, which is good, but the fact that it fails when the NumSets are set equal to the same object instance rather than when their sets are the same this tells you what the `==` in your `Equals` method is _actually_ doing. )

Assignment4/Assignment4.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            Assert.IsTrue(one == two);
+            Assert.IsFalse(one == three);
+        }
+
+        [TestMethod]
+        public void NotEqualsOperator_ReturnsValidValue()
+        {
+            NumSet one = new(1, 2, 3, 5, 6);
+            NumSet two = new(1, 2, 3, 5, 6);
```
mmwoodfo, line 86:
```suggestion
```

Assignment4/Assignment4.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,100 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+
+namespace Assignment4.Tests
+{
+    [TestClass]
+    public class NumSetTests
```
mmwoodfo, line 8: It would be good to have a test that confirms when you modify the returned `array` it does not modify the `HashSet`

PR #50, opened 2021-02-02
Assignment4

jschofield1:
Collaborated with @coproduct

### DIFF
--- a/Assignment4/MyLibrary/MyLibrary.Tests/NumSetTests.cs
+++ b/Assignment4/MyLibrary/MyLibrary.Tests/NumSetTests.cs
@@ -205,5 +205,35 @@ namespace MyLibrary.Tests
 
             //Assert
         }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ImplicitConvertArray_TakesNullHashSet_ThrowsArgumentNullException()
+        {
+            // Assign
+            NumSet numSet = new NumSet()
+            {
+                MySet = null
+            };
+
+            // Act
+            int[] array = numSet;
+
+            // Assert
+        }
+
+        [TestMethod]
+        public void ImplicitConvertArray_TakesNumSet_HasSameValues()
+        {
+            // Assign
+            int[] array = new int[] { 1, 2, 3, 4, 5 };
+            NumSet numSet = new NumSet(array);
+
+            // Act
+            int[] returnArray = numSet;
+
+            // Assert
+            CollectionAssert.AreEqual(array, returnArray);
+        }
     }
 }
--- a/Assignment4/MyLibrary/MyLibrary.Tests/NumSetTests.cs
+++ b/Assignment4/MyLibrary/MyLibrary.Tests/NumSetTests.cs
@@ -1,0 +1,209 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+
+namespace MyLibrary.Tests
+{
+    [TestClass]
+    public class NumSetTests
+    {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void NumSet_PassesNullArgumentsToConstructor_ThrowsArgumentNullException()
+        {
+            //Assign
+
+            //Act
+            _ = new NumSet(null!);
+
+            //Assert
+        }
+
+        [TestMethod]
+        public void NumSet_PassesArrayToConstructor_AssignsArrayToHashSet()
+        {
+            //Assign
+            int[] array = new int[] { 1, 2, 3, 4, 5 };
+            NumSet numSet = new NumSet(array);
+
+            //Act
+            _ = new HashSet<int>(array);
+
+            //Assert
+            Assert.IsNotNull(numSet.MySet);
+        }
+
+        [TestMethod]
+        public void ToString_TakesHashSet_ReturnsStringOfHashSet()
+        {
+            //Assign
+            int[] array = new int[] { 1, 2, 3, 4, 5 };
+            NumSet numSet = new NumSet(array);
+
+            //Act
+            string arrayToString = numSet.ToString();
+
+            //Assert
+            Assert.AreEqual(arrayToString, "1, 2, 3, 4, 5");
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ToString_TakesNullHashSet_ThrowsArgumentNullException()
+        {
+            //Assign
+            NumSet? numSet = new NumSet
+            {
+                MySet = null
+            };
+
+            //Act
+            _ = numSet.ToString();
+
+            //Assert
+        }
+
+        [TestMethod]
+        public void Equals_PassesDifferentNumSets_ReturnsFalse()
+        {
+            //Assign
+            int[] array1 = new int[] { 1, 2, 3, 4, 5 };
+            int[] array2 = new int[] { 6, 7, 8, 9, 10 };
+            NumSet numSet1 = new NumSet(array1);
+            NumSet numSet2 = new NumSet(array2);
+
+            //Act
+
+            //Assert
+            Assert.IsFalse(numSet1.Equals(numSet2));
+        }
+
+        [TestMethod]
+        public void Equals_PassesTheSameNumSet_ReturnsTrue()
+        {
+            //Assign
+            int[] array = new int[] { 1, 2, 3, 4, 5 };
+            NumSet numSet = new NumSet(array);
+
+            //Act
+
+            //Assert
+            Assert.IsTrue(numSet.Equals(numSet));
+        }
+
+        [TestMethod]
+        public void Equals_PassesNonNumSetObject_ReturnsFalse()
+        {
+            //Assign
+            int[] array = new int[] { 1, 2, 3, 4, 5 };
+            NumSet numSet = new NumSet(array);
+            object obj = new object();
+
+            //Act
+
+            //Assert
+            Assert.IsFalse(numSet.Equals(obj));
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Equals_PassesNullArray_ThrowsArgumentNullException()
+        {
+            //Assign
+            int[] array1 = new int[] { 1, 2, 3, 4, 5 };
+            int[] array2 = new int[] { 6, 7, 8, 9, 10 };
+            NumSet numSet1 = new NumSet(array1);
+            NumSet numSet2 = new NumSet(array2);
+            numSet1.MySet = null;
+
+            //Act
+            _ = numSet1.Equals(numSet2);
+
+            //Assert
+        }
+
+        [TestMethod]
+        public void GetHashCode_TakesNumSet_ReturnsCorrectHashCode()
+        {
+            //Assign
+            int[] array = new int[] { 1, 2, 3, 4, 5 };
+            NumSet numSet1 = new NumSet(array);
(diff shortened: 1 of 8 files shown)


jschofield1:
Collaborated with @coproduct

    Both projects created with appropriate properties: ✔
    NumSet constructor accepts variable arguments: ✔
    NumSet ToString implemented correctly: ✔
    NumSet equality operators implemented correctly: ✔
    NumSet fully unit tested: ✔
    SeWriter created in nested namespace: ✔
    SeWriter writes set to file: ✔
    SeWriter properly implements IDisposable: ✔
    SeWriter fully unit tested: ✔

Extra Credit

NumSet contains cast operator that is unit tested: ❌ 
Projects updated to properly use Directory.Build.props: ❌

coproduct:
Extra Credit

NumSet contains cast operator that is unit tested: :heavy_check_mark: 
Projects updated to properly use Directory.Build.props: :heavy_check_mark:

Assignment4/MyLibrary/MyLibrary/NumSet.cs
```diff
@@ -0,0 +1,94 @@
...
+    public class NumSet
+    {
+        private HashSet<int>? mySet;
+
+        public HashSet<int>? MySet 
+        { 
+            get => mySet; 
+            set => mySet = value ??
+                throw new ArgumentNullException(nameof(value)); 
+        }
```
mmwoodfo, line 16: You don't need this - you can just have a private `HashSet` variable.

Assignment4/MyLibrary/MyLibrary/NumSet.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace MyLibrary
+{
+    public class NumSet
+    {
+        private HashSet<int>? mySet;
```
mmwoodfo, line 9:
```suggestion
        private readonly HashSet<int> _MySet;
```

Assignment4/MyLibrary/MyLibrary/NumSet.cs
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        public NumSet(params int[] array)
+        {
+            mySet = new HashSet<int>(array);
+        }
+
+        public override string ToString()
+        {
+            return String.Join(", ", MySet!);
```
mmwoodfo, line 25:
```suggestion
            return string.Join(", ", MySet);
```
Use the C# keywords rather than BCL names.

Assignment4/MyLibrary/MyLibrary/NumSet.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public override bool Equals(object? obj)
+        {
+            if (ReferenceEquals(this, obj))
+                return true;
+
+            if (obj is not NumSet)
+                return false;
+
+            return MySet == ((NumSet)obj).MySet;
```
mmwoodfo, line 36: You should use `SetEquals` here rather than `==`. The `==` operator will only return true if the two objects are the same instance, it will fail for two separately instantiated objects that have the same hash set.

Assignment4/MyLibrary/MyLibrary/SetWriter.cs
```diff
@@ -0,0 +1,84 @@
...
+        private bool _DisposedValue;
+
+        private string? _filePath;
+
+        public string? FilePath
+        {
+            get => _filePath;
+            set => _filePath = value ??
+                throw new ArgumentNullException(nameof(value));
+        }
```
mmwoodfo, line 17: You don't need to save the file path in a variable.

Assignment4/MyLibrary/MyLibrary/SetWriter.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        private StreamWriter? fileWriter;
+
+        public StreamWriter? FileWriter
+        {
+            get => fileWriter;
+            set => fileWriter = value ?? 
+                throw new ArgumentNullException(nameof(value));
+        }
```
mmwoodfo, line 26:
```suggestion
        private readonly StreamWriter _FileWriter;
```
You only need a readonly private variable here. It doesn't have to be marked as nullable because you should always be creating this in the constructor.

Assignment4/MyLibrary/MyLibrary/SetWriter.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        public SetWriter(string filePath)
+        {
+            FileWriter = new StreamWriter(filePath);
+        }
+
+        public void WriteSet(NumSet numSet)
+        {
+            FileWriter!.Write(numSet.ToString());
```
mmwoodfo, line 35:
```suggestion
            FileWriter.Write(numSet.ToString());
```
Use the `!` cautiously. If you find that you are using it often there is likely something wrong with how you are writing your code. 

Also, you may consider using `WriteLine` here.

Assignment4/MyLibrary/MyLibrary/SetWriter.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System;
+using System.IO;
+
+namespace MyLibrary.Writer
```
mmwoodfo, line 4: 👍🏻

Assignment4/MyLibrary/MyLibrary.Tests/SetWriterTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            //Assign
+
+            //Act
+            _ = new SetWriter(null!);
+
+            //Assert
+        }
+
+        [TestMethod]
+        public void WriteSet_WritesNumSetToFile_NumSetOnFileMatches()
```
mmwoodfo, line 37: You should dispose of the file you've created for this test. You'll get an I/O exception from `GetTempFileName` if it can't generate a unique file name, causing your test to fail unexpectedly. The solution to this is to clean up your temporary files.

Assignment4/MyLibrary/MyLibrary.Tests/SetWriterTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void WriteSet_PassesNullArgument_ThrowsArgumentNullException()
+        {
+            //Assign
+            string filePath = Path.GetTempFileName();
+            SetWriter setWriter = new SetWriter(filePath);
+            NumSet numSet = new NumSet
+            {
+                MySet = null
+            };
```
mmwoodfo, line 63: You should not be able to create the `NumSet` like this. `MySet` should be a private readonly value that only gets set through the constructor of the `NumSet` class.

Assignment4/MyLibrary/MyLibrary.Tests/SetWriterTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        public void WriteSet_TakesNullFileWriter_ThrowsArgumentNullException()
+        {
+            //Assign
+            string filePath = Path.GetTempFileName();
+            int[] testArray = new int[] { 1, 2, 3, 4, 5 };
+            SetWriter setWriter = new SetWriter(filePath);
+            NumSet numSet = new NumSet(testArray);
+            
+            //Act
+            setWriter.FileWriter = null;
```
mmwoodfo, line 82: This should not be something that should happen if you check for a null file name in the constructor create a new `StreamWriter` when that null check passes. The `StreamWriter` should not be accessible or modifiable from outside of the class like this.

Assignment4/MyLibrary/MyLibrary.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,239 @@
...
+            //Assert
+        }
+
+        [TestMethod]
+        public void GetHashCode_TakesNumSet_ReturnsCorrectHashCode()
+        {
+            //Assign
+            int[] array = new int[] { 1, 2, 3, 4, 5 };
+            NumSet numSet1 = new NumSet(array);
+            NumSet numSet2 = numSet1;
```
mmwoodfo, line 131:
```suggestion
            int[] array = new int[] { 1, 2, 3, 4, 5 };
            NumSet numSet1 = new NumSet(array);
            NumSet numSet2 = new NumSet(array);
```
You should create two separate objects, rather than setting the second `NumSet` equal to the first one.

Assignment4/MyLibrary/MyLibrary.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,239 @@
...
+        public void EqualityOperators_ComparesEqualAndUnequalNumSets_ReturnsExpectedBools()
+        {
+            //Assign
+            int[] array1 = new int[] { 1, 2, 3, 4, 5 };
+            int[] array2 = new int[] { 6, 7, 8, 9, 10 };
+            
+            NumSet numSet1 = new NumSet(array1);
+            NumSet numSet2 = new NumSet(array2);
+            NumSet? numSet3 = null;
+            NumSet? numSet4 = null;
```
mmwoodfo, line 167:
```suggestion
            NumSet numSet1 = new NumSet(array1);
            NumSet numSet2 = new NumSet(array2);
            NumSet? numSet3 = null;
            NumSet? numSet4 = null;
            NumSet numSet5 = new NumSet(array2);
```

Assignment4/MyLibrary/MyLibrary.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            //Act
+
+            //Assert
+            Assert.IsTrue(numSet3! == numSet4!);
+            Assert.IsTrue(numSet1 != numSet2);
+            Assert.IsFalse(null! == numSet1);
+            Assert.IsFalse(numSet1! == null!);
```
mmwoodfo, line 175:
```suggestion
            Assert.IsTrue(numSet3! == numSet4!);
            Assert.IsTrue(numSet1 != numSet2);
            Assert.IsFalse(null! == numSet1);
            Assert.IsFalse(numSet1! == null!);
            Assert.IsTrue(numSet2 == numSet5); // This test will cause your code to fail
```
This should not fail - but it will because you are using `==` in your override of `Equals`.

Assignment4/MyLibrary/MyLibrary.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        [TestMethod]
+        public void ReturnArray_TakesHashSet_ReturnsIntArray()
```
mmwoodfo, line 179: It would be good to have a test to check if modification of the array modifies the hash set.

Assignment4/MyLibrary/MyLibrary.Tests/NumSetTests.cs
```diff
@@ -0,0 +1,239 @@
...
+        public void Equals_PassesTheSameNumSet_ReturnsTrue()
+        {
+            //Assign
+            int[] array = new int[] { 1, 2, 3, 4, 5 };
+            NumSet numSet = new NumSet(array);
+
+            //Act
+
+            //Assert
+            Assert.IsTrue(numSet.Equals(numSet));
```
mmwoodfo, line 91:
```suggestion
            //Assign
            int[] array = new int[] { 1, 2, 3, 4, 5 };
            NumSet numSet1 = new NumSet(array);
            NumSet numSet2 = new NumSet(array);

            //Act

            //Assert
            Assert.IsTrue(numSet1.Equals(numSet2)); // This should not fail
```
This will fail because of the `==` being used to compare in your override of `Equals`.

PR #54, opened 2021-02-08
Assignment5

NPThompson:
no extra credit

### DIFF
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -6,8 +6,9 @@ namespace GenericsHomework.Tests
     public class NodeTests
     {
         [TestMethod]
-        public void FooTest()
+        public void Node_ConstructorIsGeneric()
         {
+            Node<int> intNode = new(10);
         }
     }
 }
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -5,10 +5,36 @@ namespace GenericsHomework.Tests
     [TestClass]
     public class NodeTests
     {
+        private (Node<int>,Node<string>) makeNodes()
+        {
+            return (new Node<int>(10), new Node<string>("Miracle Max"));
+        }
+
         [TestMethod]
         public void Node_ConstructorIsGeneric()
         {
-            Node<int> intNode = new(10);
+            var (intNode, stringNode) = makeNodes();
+        }
+
+        [TestMethod]
+        public void Node_ToStringWorks()
+        {
+            var (intNode, stringNode) = makeNodes();
+
+            Assert.AreEqual<string>(intNode.ToString(), "10");
+            Assert.AreEqual<string>(stringNode.ToString(), "Miracle Max");
+        }
+
+        [TestMethod]
+       public void Node_Insert_AltersNode()
+        {
+            Node<int> node0 = new(0);
+            Node<int> node1 = new(1);
+
+            node0.Insert(node1);
+            node1.Insert(node0);
+
+            Assert.AreNotEqual<int>(node0.Next.data, node1.Next.data);
         }
     }
 }
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -32,9 +32,34 @@ namespace GenericsHomework.Tests
             Node<int> node1 = new(1);
 
             node0.Insert(node1);
-            node1.Insert(node0);
-
+         
             Assert.AreNotEqual<int>(node0.Next.data, node1.Next.data);
+        }
+
+        [TestMethod]
+        public void Node_IsCircular()
+        {
+            Node<int> node0 = new(0);
+            
+            node0.Insert(new(1));
+            node0.Insert(new(2));
+
+            Assert.AreNotEqual<Node<int>>(node0.Next, node0);
+            Assert.AreNotEqual<Node<int>>(node0.Next.Next, node0);
+            Assert.AreEqual<Node<int>>(node0.Next.Next.Next, node0);
+        }
+
+        // see Node.cs for comment
+        [TestMethod]
+        public void Node_Clear_RemovesAllOtherNodes()
+        {
+            Node<int> node0 = new(0);
+            node0.Insert(new(1));
+            node0.Insert(new(2));
+            node0.Insert(new(3));
+
+            node0.Clear();
+            Assert.AreEqual<Node<int>>(node0, node0.Next);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


mmwoodfo:
@NPThompson When doing your code reviews it is more beneficial for the person you are reviewing to give them inline comments (similar to what you see me doing on your PR) rather than a single comment listing all of the issues. This isn't something you are going to be dinged for on this assignment, but you may lose points on future assignments for unclear or difficult to read reviews.'

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+
+namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        public Node(T init) 
+        {
+            this.data = init;
+            this.Next = this;
```
mmwoodfo, line 10:
```suggestion
            _Data = init;
            Next = this;
```
You don't need to use `this` here.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        public Node(T init) 
+        {
+            this.data = init;
+            this.Next = this;
+        }
+        
+        public Node()
+        {
+            this.Next = this;
```
mmwoodfo, line 15:
```suggestion
            Next = this;
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,43 @@
...
+        public T? data;
+        public Node<T> Next { get; private set; }
+
+
+        public override string ToString()
+        {
+            return this.data?.ToString() ?? "";
+        }
+
+        public void Insert(Node<T> newNext)
```
mmwoodfo, line 27: A new node should be created in the Insert property, the parameter here should just be the value. The insertion of the node should also be handled in the `Next` property, not in the Insert method. The insert method should simply invoke the `Next` property.

> Add an Insert method that takes a value and inserts a new Node instance after the current node (by invoking the Next property).

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,43 @@
...
+            this.data = init;
+            this.Next = this;
+        }
+        
+        public Node()
+        {
+            this.Next = this;
+        }
+
+        public T? data;
```
mmwoodfo, line 18:
```suggestion
        public T _Data;
```
This does not need to be marked as nullable.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,43 @@
...
+        public Node<T> Next { get; private set; }
+
+
+        public override string ToString()
+        {
+            return this.data?.ToString() ?? "";
```
mmwoodfo, line 24:
```suggestion
            return _Data?.ToString() ?? "";
```

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,43 @@
...
+            this.Next = this;
+        }
+        
+        public Node()
+        {
+            this.Next = this;
+        }
+
+        public T? data;
+        public Node<T> Next { get; private set; }
```
mmwoodfo, line 19:
```suggestion
        public Node<T> Next { get; private set; }
```
This shouldn't be an auto property, this should be a property with its `get` and `set` implemented.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,43 @@
...
+        public void Insert(Node<T> newNext)
+        {
+            newNext.Next = this.Next;
+            this.Next = newNext;
+        }
+
+        public void Clear()
+        {
+            // This is all that is necessary
+            this.Next = this;
```
mmwoodfo, line 36:
```suggestion
            Next = this;
```

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,65 @@
...
+            var (intNode, stringNode) = makeNodes();
+        }
+
+        [TestMethod]
+        public void Node_ToStringWorks()
+        {
+            var (intNode, stringNode) = makeNodes();
+
+            Assert.AreEqual<string>(intNode.ToString(), "10");
+            Assert.AreEqual<string>(stringNode.ToString(), "Miracle Max");
```
mmwoodfo, line 25:
```suggestion
            Assert.AreEqual(intNode.ToString(), "10");
            Assert.AreEqual(stringNode.ToString(), "Miracle Max");
```
Not needed. This is true for all of the `Asserts` that do this.

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,65 @@
...
+            Node<int> node0 = new(0);
+            
+            node0.Insert(new(1));
+            node0.Insert(new(2));
+
+            Assert.AreNotEqual<Node<int>>(node0.Next, node0);
+            Assert.AreNotEqual<Node<int>>(node0.Next.Next, node0);
+            Assert.AreEqual<Node<int>>(node0.Next.Next.Next, node0);
+        }
+
```
mmwoodfo, line 51:
```suggestion
        [TestMethod]
       public void Node_Insert_MultipleNodes()
        {
            // 1 -> 2 -> 3 -> 4
            // ┖──────────────┘

            Node<int> node1 = new(1);
            Node<int> node2 = new(2);
            Node<int> node3 = new(3);
            Node<int> node4 = new(4);

            node1.Insert(node2);
            node2.Insert(node4);
            node2.Insert(node3);
         
            Assert.AreEqual(node1, node3.Next.Next);
            Assert.AreEqual(node2.Next, node3);
            Assert.AreEqual(node1.Next, node2);
            Assert.AreEqual(node3.Next, node4);
        }
```

It would be good to have a few other tests that test for the robustness of your linked list. Making sure that inserting more than two nodes works & the order of insertion doesn't mess up your list. The above code will pass on your current implementation.

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        // see Node.cs for comment
+        [TestMethod]
+        public void Node_Clear_RemovesAllOtherNodes()
```
mmwoodfo, line 54: You should consider adding a test that tests if the nodes really did get cleaned up by the garbage collector. 

You can do this by creating the nodes you want to be cleaned up in an isolated scope & adding a weak reference to them. After exiting that scope, call clear on the node you plan to not clean up followed by a call to `GC.Collect()` and `GC.WaitForPendingFinalizers()`. Then check if the weak references are null or not.

mmwoodfo:
Code review was done after the due date.

PR #55, opened 2021-02-09
Assignment5

cmartinezaraujo:
Assignment 5 complete. Extra credit not attempted.


Sithale (commented):
Define the Node class
Extra Credit

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ❌✔
Implement Systm.Collections.Generic.ICollection<T> on the Node class ❌✔
Fundamentals

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
...
+        private Node<T>? _Next;
+
+        public T Value { get => _Value!; private  set => _Value = value;}
+
+        public Node<T> Next
+        {
+            get => _Next!;
+            set
+            {
+                value._Next = this;
```
cmartinezaraujo, line 17: Found typo when using this class for the last part of assignment 6. Should be value.next = this.Next;

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+
+namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        private T? _Value;
+        private Node<T>? _Next;
```
mmwoodfo, line 8:
```suggestion
        private T _Value;
        private Node<T> _Next;
```
You do not need to set these as nullable. If you set them in the constructor you won't have any warnings.

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+
+namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        private T? _Value;
+        private Node<T>? _Next;
+
+        public T Value { get => _Value!; private  set => _Value = value;}
```
mmwoodfo, line 10:
```suggestion
```
You can just use the private variable.

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                value._Next = this;
+                _Next = value ?? throw new ArgumentNullException(nameof(value));
+            }
+        }
+
+        public Node(T value)
+        {
+            Value = value;
+            Next = this;
```
mmwoodfo, line 25:
```suggestion
            _Value = value;
            _Next = this;
```
Use the private variables here to avoid marking them as nullable.

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public override string ToString()
+        {
+            if(Value is null)
+            {
+                throw new ArgumentNullException(nameof(Value));
+            }
```
mmwoodfo, line 33: Throwing is likely not what you want here.

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            return Value.ToString()!;
```
mmwoodfo, line 35:
```suggestion
            return _Value?.ToString() ?? "";
```

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public void Insert(T value)
+        {
+            Node<T> newNode = new Node<T>(value);
+            this.Next = newNode;
```
mmwoodfo, line 41:
```suggestion
            Next = newNode;
```
No need to use `this` here.

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public void Clear()
+        {
+            Next = this;
```
mmwoodfo, line 46: 👍🏻

Assignment5/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            Node<int> node = new Node<int>(25);
+
+            Node<int> differentNodeReference = new Node<int>(25);
+
+            Assert.AreNotEqual<Node<int>>(node, differentNodeReference);
+        }
+
+        [TestMethod]
+
+        public void NodeInsertsNewNodeToListAndLoopsBack()
```
mmwoodfo, line 65: You should have a test with more than just two nodes. While this test does pass, if you were to test the following code:
```C#
// 1 -> 2 -> 3
// ┖────┘

// Arrange

// Act
one.Next.Insert(2);
var two = one.Next;

two.Next.Insert(3);
var three = two.Next;

```

```
1 -> 2 -> 3
┖─────────┘
```

You'll have this:
```
1 -> 3
┖────┘

2 -> 1 -> 3
     ┖────┘
```

Assignment5/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            Assert.AreEqual<string>("StringOne", node.ToString());
+            Assert.AreEqual<string>("StringTwo", node.Next.ToString());
+            Assert.AreEqual<string>("StringOne", node.Next.Next.ToString());
+
+        }
+
+        [TestMethod]
+
+        public void CallingClearOnCurrentNodeClearsAllOtherNodes()
```
mmwoodfo, line 79: You may also want to verify that the garbage collector actually cleaned up the detached nodes. You can do this by creating weak references to the nodes that will get cleaned up and then get them out of scope before you call `GC.Collect()` and `GC.WaitForPendingFinalizers()`.

Assignment5/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        [TestMethod]
+
+        public void NodeInsertsNewNodeToListAndLoopsBack()
+        {
+            Node<string> node = new Node<string>("StringOne");
+
+            node.Insert("StringTwo");
+
+            Assert.AreEqual<string>("StringOne", node.ToString());
```
mmwoodfo, line 71: Avoid using other parts of your code if you aren't actively testing them. Unit tests should test individual _units_ of your code.

Assignment5/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            node.Insert("Two");
+            node.Insert("Three");
+            node.Insert("Four");
+            node.Insert("Five");
+
+            node.Clear();
+
+            Assert.AreEqual<string>("One", node.Value);
+            Assert.AreEqual<string>("One", node.Next.Value);
```
mmwoodfo, line 91:
```suggestion
            Assert.AreEqual("One", node.Value);
            Assert.AreEqual("One", node.Next.Value);
```
No needed. This goes for all of the above tests doing this as well.

PR #56, opened 2021-02-09
Assignment5

MeguminXIII:


NPThompson:
I will review this

NPThompson:
        Define the Node class
            The Next property setter should be private. ❌

Looks good. The Next property setter is public in your implementation, tho. The fix is simple.

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        [TestMethod]
+        public void Node_TestingInsert_ListShouldStillBeCircular()
+        {
+            Node<string> one = new("Inigo_Montoya");
+            one.Insert("Red");
+
+            Assert.AreEqual<string>("Red", one.Next.ToString());
+            Assert.AreEqual<string>("Inigo_Montoya", one.Next.Next.ToString());
+            Assert.AreEqual<string>("Red", one.Next.Next.Next.ToString());
```
mmwoodfo, line 74: Avoid using other parts of your code if you aren't actively testing them. Unit tests should test individual _units_ of your code.

mmwoodfo:
@MeguminXIII When doing your code reviews it is more beneficial for the person you are reviewing to give them inline comments (similar to what you see me doing on your PR) rather than a single comment listing all of the issues. This isn't something you are going to be dinged for on this assignment, but you may lose points on future assignments for unclear or difficult to read reviews.

PR #57, opened 2021-02-09
Assignment5

alexvthornton:
Assignment 5. 
partner: @kaykay38
 Extra credit #2 attempted


alexvthornton:
Define the Node class
Whether it is sufficient to only set Next to itself✔
Extra Credit

Implement Systm.Collections.Generic.ICollection<T> on the Node class ✔
Fundamentals

kaykay38:
# EWU-CSCD371-2021-Winter

## See [Docs](Docs)

## Assignment 5

The purpose of this assignment is to write a class library is to learn how to write a generic class and a generic method. To accomplish this you will write a linked list that circles back on itself.

## Reading

Catch Up On Reading (Chapter 1-15 except Chapter 14). This is your chance to get caught up - use it!

## Instructions

- Create a *class library* project called "GenericsHomework.". ✔
- Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item.
  - Define the `Node` class
  - Include a constuctor that takes a value.  (No validation is necessary on the value). ✔
  - Add a `ToString()` override that writes out the value's `ToString()` result. ✔
  - Add a `Next` property that references the next node or else refers back to itself if there are no other nodes in the list. ✔
    - The `Next` property should be non-nullable (careful to follow the non-nullable property guidelines) ✔
    - The `Next` property setter should be private. ✔
    - In addition to non-null validation, the body of the `Next` property should insert the next `Node` (the `value`) into the list. ✔
  - Add an `Insert` method that takes a value and inserts a new `Node` instance after the current node (by invoking the `Next` property). ✔
  - Add a clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be
  concerned with the following:
    - Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage colleciton because all the items point to each other and therefore may never be garbage collected. ✔

## Extra Credit

1. Implement a `VennDiagram` structure that contains `n` `Circle`s that only contains homogenous **reference types** of any type. ❌

- Each circle contains n items and each item can belong to one more more `Circle` instances.
- You are not required to use a `Node` from earlier in the homework for your venn diagram implementation.
- You are welcome to use exising BCL generic classes for the extra credit.

2. Implement `Systm.Collections.Generic.ICollection<T>` on the `Node` class ✔

## Fundamentals

- Ensure you enable nullable reference types, net5 targetted, C# 9.0, warnings as errors, and enabled .NET analyzers for both projects ✔
  - For this assignment, always use `Assert.AreEquals<T>()` (the generic version)
- All of the above should be unit tested ✔

## Code Coverage
| Module                    | Line    | Branch | Method |
|-----------------------|--------|---------|----------|
| GenericsHomework | 94.3% | 76.08% | 94.44%  |

|               | Line    | Branch  | Method |
|----------|--------|----------|----------|
| Total      | 94.3% | 76.08% | 94.44%  |
| Average | 94.3% | 76.08% | 94.44%  |

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,358 @@
...
+namespace GenericsHomework.Tests
+{
+
+    [TestClass]
+    public class NodeTests
+    {
+
+        public string ToString(int[] array) => "{" + string.Join(", ", array) + "}";
+        public string ToString(ArrayList list) => "{" + string.Join(", ", list.ToArray()) + "}";
+        public string ToString<T>(List<T> list) => "{" + string.Join(", ", list.ToArray()) + "}";
```
mmwoodfo, line 15: These can all be static.
```suggestion
        public static string ToString(int[] array) => "{" + string.Join(", ", array) + "}";
        public static string ToString(ArrayList list) => "{" + string.Join(", ", list.ToArray()) + "}";
        public static string ToString<T>(List<T> list) => "{" + string.Join(", ", list.ToArray()) + "}";
```

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,358 @@
...
+
+
+        public Node<int> CreateInt4NodeList(int num1, int num2, int num3, int num4)
```
mmwoodfo, line 18:
```suggestion
        public static Node<int> CreateInt4NodeList(int num1, int num2, int num3, int num4)
```
Also can be marked as static, the same goes for the methods below.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        [TestMethod]
+        public void Add_Given4StringNodes_EqualToExpectedString()
+        {
+            Node<string> list = new("First");
+            list.Add("Second");
+            list.Add("Third");
+            list.Add("Fourth");
+            string actual = $"{list.ToString()} {list.Next.ToString()} {list.Next.Next.ToString()} {list.Next.Next.Next.ToString()} {list.Next.Next.Next.Next.ToString()}";
```
mmwoodfo, line 67:
```suggestion
            string actual = $"{list} {list.Next} {list.Next.Next} {list.Next.Next.Next} {list.Next.Next.Next.Next}"; 
```
`ToString()` will be implicitly called when you are doing string interpolation.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,358 @@
...
+            Node<int> node1 = new(789);
+
+            Assert.AreEqual("789", node1.Data.ToString());
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void NewInstance_GivenNullValue_ThrowArgumentNullException()
+        {
+            Node<string> node1 = new(null!);
```
mmwoodfo, line 57:
```suggestion
            new Node<string>(null!);
```

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,358 @@
...
+
+        [TestMethod]
+        public void Clear_Given4StringNodesClearFromNode3_DeferencedNode1NextReference()
+        {
+            Node<string> node1 = new Node<string>("First");
+            Node<string> node2 = node1.Insert("Second");
+            Node<string> node3 = node2.Insert("Third");
+            Node<string> node4 = node3.Insert("Fourth");
+            node3.Clear();
+            Assert.AreEqual<string>("First", node1.Next.Next.ToString()!);
```
mmwoodfo, line 89:
```suggestion
            Assert.AreEqual("First", node1.Next.Next.ToString()!);
```
Not needed. This goes for the rest of the `Assert` declerations that do this.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,358 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace GenericsHomework.Tests
+{
+
+    [TestClass]
+    public class NodeTests
```
mmwoodfo, line 10: A general rule of thumb, your unit tests should be testing the individual _units_ of your code. Avoid combining multiple parts of your code into a single unit test if you can avoid it. Most of your test methods are using your `ToString` override to check the validity of non-related components in your code. If your `ToString` method ends up getting changed to where it now does not work as expected - all of your unit tests will fail and it will be much harder to determine the root cause of the issue. 

With that said - the majority of these test cases should be refactored to NOT use the `ToString` method to check correctness. _(not that you are required to refactor.)_

GenericToString/GenericToString.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace GenericToString
+{
+    public static class GenericToString
+    {
+        public static string ToString<T>(this List<T> list) => "{" + string.Join(", ", list.ToArray()) + "}";
+        public static string ToString<T>(this ArrayList list) => "{" + string.Join(", ", list.ToArray()) + "}";
+
+    }
+}
```
mmwoodfo, line 12: Is this leftover code?

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,203 @@
...
+                cur = cur.Next;
+                first = false;
+            }
+            return false;
+        }
+
+        public void CopyTo(T[] array, int arrayIndex)
+        {
+            if (array == null)
+                throw new ArgumentNullException("The array cannot be null.");
```
mmwoodfo, line 141:
```suggestion
                throw new ArgumentNullException(nameof(array));

```

PR #58, opened 2021-02-09
Assignment5

Zophed:


worseTyler:
I will review this

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+                _Value = value ?? throw new ArgumentNullException(nameof(value));
+            }
+        }
+        public Node<T> Next
+        {
+            get => _Next!;
+
+            private set
+            {
+                value._Next = this;
```
worseTyler, line 25: I don't think you need this line

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+                _Next = value ?? throw new ArgumentNullException(nameof(value));
+            }
+        }
+
+        public Node(T value)
+        {
+            this.Value = value;
+            this.Next = this;
```
worseTyler, line 33: Remove the this prefix, also on line 39

GenericsHomework/GenericsHomework/Node.cs
worseTyler, line 25: This makes the list circular, gotcha

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            //Act
+
+            //Assert
+            Assert.AreEqual<string>("42", intNode.ToString());
+            Assert.AreEqual<string>("The Dread Pirate Roberts", stringNode.ToString());
+        }
+
+        [TestMethod]
+        public void NextProperty_ReturnsCorrectNode()
```
worseTyler, line 41: Does this test the next property?

worseTyler:
Instructions
Define the Node class
Extra Credit

Implement Systm.Collections.Generic.ICollection<T> on the Node class ❌
Fundamentals
All of the above should be unit tested ✔l

mmwoodfo:
When doing your code reviews it is more beneficial for the person you are reviewing to give them inline comments (similar to what @worseTyler did on this PR) rather than a single comment listing all of the issues. This isn't something you are going to be dinged for on this assignment, but you may lose points on future assignments for unclear or difficult to read reviews.'

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+
+namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        private T? _Value;
+        private Node<T>? _Next;
```
mmwoodfo, line 8: If you set these in the constructor you do not have to make them as nullable.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            this.Next = this;
+        }
+
+        public override string ToString()
+        {
+            if (Value is null)
+            {
+                throw new ArgumentNullException(nameof(Value));
+            }
```
mmwoodfo, line 58: The code below will return an empty string if `Value` is null, so you don't need this here.

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            //Act
+            stringNode.Clear();
+
+            //Assert
+            Assert.IsTrue(stringNode.ToString() == stringNode.Next.ToString());
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ToString_PassedNullNode_ThrowsException()
```
mmwoodfo, line 83: This test name is misleading.

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            //Assign
+            Node<int> headNode = new Node<int>(42);
+
+            //Act
+            headNode.Insert(43);
+
+            //Assert
+            Assert.AreEqual<string>("42", headNode.ToString());
+            Assert.AreEqual<string>("43", headNode.Next.ToString());
+            Assert.AreEqual<string>("42", headNode.Next.Next.ToString());
```
mmwoodfo, line 65: Try to avoid using your `ToString` method when unit testing unrelated code. Your unit tests should be focused on a single _unit_ of your code, in this case, just the `Insert` method. Another way to do this below:
```suggestion
            Node<int> secondNode = headNode.Next;

            Assert.AreEqual(secondNode, headNode.Next);
            Assert.AreNotEqual(secondNode, headNode);
            Assert.AreEqual(headNode, headNode.Next.Next);
```

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void ToString_ReturnsCorrectString()
+        {
+            //Assign
+            var (stringNode, intNode) = TestNodes();
+
+            //Act
+
+            //Assert
+            Assert.AreEqual<string>("42", intNode.ToString());
+            Assert.AreEqual<string>("The Dread Pirate Roberts", stringNode.ToString());
```
mmwoodfo, line 37:
```suggestion
            Assert.AreEqual("42", intNode.ToString());
            Assert.AreEqual("The Dread Pirate Roberts", stringNode.ToString());
```

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            Node<string> secondNode = new Node<string>("Princess Buttercup");
+
+            //Act
+
+            //Assert
+            Assert.AreNotEqual<Node<string>>(firstNode, secondNode);
+        }
+
+        [TestMethod]
+        public void InsertMethodGivenNewNode_AssignsNextAndLoopsBackToNode()
```
mmwoodfo, line 54: You should test more than two nodes being inserted into the list. You should also test the order in which the nodes get added to the list. When writing unit tests the goal is to write unit tests that stress test your code. Hit the limits, test edge cases, and hopefully reveal bugs.

GenericsHomework/GenericsHomework/Node.cs
mmwoodfo: @worseTyler @Zophed
Not quite. While this works with two nodes in the list, once you add three you will start to get _detached head and neck_ like structure in your list.

The following unit test will cause a failure in your code.
```C#
// 1 -> 2 -> 3
// ┖────┘

// Arrange

// Act
one.Next.Insert(2);
var two = one.Next;

two.Next.Insert(3);
var three = two.Next;

```

```
1 -> 2 -> 3
┖─────────┘
```

You'll have this:
```
1 -> 3
┖────┘

2 -> 1 -> 3
     ┖────┘
```

There are easy ways within visual studio to look at your variables and data structures. [Here are some tips for that](https://devblogs.microsoft.com/devops/7-ways-to-look-at-the-values-of-variables-while-debugging-in-visual-studio/).

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            headNode.Insert(43);
+
+            //Assert
+            Assert.AreEqual<string>("42", headNode.ToString());
+            Assert.AreEqual<string>("43", headNode.Next.ToString());
+            Assert.AreEqual<string>("42", headNode.Next.Next.ToString());
+        }
+
+        [TestMethod]
+        public void ClearMethodRan_AllNodesExceptCurrentRemoved()
```
mmwoodfo, line 69: You could also add a test to make sure the detached nodes get collected by the garbage collector.

PR #59, opened 2021-02-09
Assignment5

Sithale:
Did not do extra credit.

### DIFF
--- a/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -1,0 +1,116 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace GenericsHomework.Tests
+{
+    [TestClass]
+    public class NodeTests
+    {
+        [TestMethod]
+        public void NodeConstructor_DifferentValuesAreCreated()
+        {
+            // Act
+            Node<double> node1 = new Node<double>(42.0);
+            Node<int> node2 = new Node<int>(13);
+            Node<string> node3 = new Node<string>("Greetings!");
+            Node<char> node4 = new Node<char>('x');
+
+            // Assert
+            Assert.AreEqual<double>(42.0, node1.Data);
+            Assert.AreEqual<int>(13, node2.Data);
+            Assert.AreEqual<string>("Greetings!", node3.Data);
+            Assert.AreEqual<char>('x', node4.Data);
+
+        }
+
+        [TestMethod]
+        public void Node_ToString_ReturnsTheCorrectString()
+        {
+            // Arrange
+            double num = 19.99;
+            string str = "Sloppy";
+
+            // Act
+            Node<double> node1 = new Node<double>(num);
+            Node<string> node2 = new Node<string>(str);
+
+            // Assert
+            Assert.AreEqual<string>(num.ToString(), node1.ToString());
+            Assert.AreEqual<string>(str.ToString(), node2.ToString());
+
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Node_ToStringPassesNull_ArgumentNullException()
+        { 
+
+            // Act
+            Node<string> node1 = new Node<string>(null!);
+
+            node1.ToString();
+            
+        }
+
+        [TestMethod]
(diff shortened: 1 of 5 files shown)


cmartinezaraujo:
I will review.

GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,50 @@
...
+        public Node<N> Next
+        {
+            get => _Next!;
+            set { value._Next = this; _Next = value; }
+
+        }
+
+        public Node(N node)
+        {
+            this.Data = node;
```
cmartinezaraujo, line 20: Remember to take out the "this" keyword in places where it Is not absolutely needed.

cmartinezaraujo (commented):
- Create a class library project called "GenericsHomework.". ✔

Define the Node class
- Include a constuctor that takes a value. (No validation is necessary on the value). ✔
- Add a ToString() override that writes out the value's ToString() result. ✔
- Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ✔
- The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ✔
- The Next property setter should be private. ✔
- In addition to non-null validation, the body of the Next property should insert the next Node (the value) into the list. ✔
- Add an Insert method that takes a value and inserts a new Node instance after the current node (by invoking the Next property). ✔

- Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage colleciton because all the items point to each other and therefore may never be garbage collected. ✔

Extra Credit:
1. Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ❌
2. Implement Systm.Collections.Generic.ICollection<T> on the Node class ❌

Fundamentals:

- Ensure you enable nullable reference types, net5 targetted, C# 9.0, warnings as errors, and enabled .NET analyzers for both projects ✔
- All of the above should be unit tested ✔

mmwoodfo:
When doing your code reviews it is more beneficial for the person you are reviewing to give them inline comments (similar to what you see me doing on your PR) rather than a single comment listing all of the issues. This isn't something you are going to be dinged for on this assignment, but you may lose points on future assignments for unclear or difficult to read reviews.'

Sithale:
Ok, thank you! I will be sure to add some more input on future reviews. I also noticed that people would start with "I will review this", and I had not done that. So, I will be doing that as well now too.
________________________________

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            double num = 19.99;
+            string str = "Sloppy";
+
+            // Act
+            Node<double> node1 = new Node<double>(num);
+            Node<string> node2 = new Node<string>(str);
+
+            // Assert
+            Assert.AreEqual<string>(num.ToString(), node1.ToString());
+            Assert.AreEqual<string>(str.ToString(), node2.ToString());
```
mmwoodfo, line 39:
```suggestion
            Assert.AreEqual(num.ToString(), node1.ToString());
            Assert.AreEqual(str.ToString(), node2.ToString());
```

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+
+            // Act
+            Node<string> node1 = new Node<string>("Howdy");
+            node1.Insert("Love");
+
+            // Assert
+            Assert.AreEqual<string>("Howdy", node1.ToString());
+            Assert.AreEqual<string>("Love", node1.Next.ToString());
+            Assert.AreEqual<string>("Howdy", node1.Next.Next.ToString());
```
mmwoodfo, line 92: Try to avoid using your `ToString` method when unit testing unrelated code. Your unit tests should be focused on a single _unit_ of your code, in this case, just the `Insert` method.

GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        }
+
+        [TestMethod]
+        public void Node_ClearFunction_AllNodesExceptFirstOneAreCleared()
```
mmwoodfo, line 97: You may consider writing a unit test to test that the garbage collector actually cleans up your detached nodes.

PR #60, opened 2021-02-09
Assignment5

worseTyler:


Zophed:
I will review this

Zophed (commented):
        Define the Node class

Extra Credit

    Implement Systm.Collections.Generic.ICollection<T> on the Node class ❌

Fundamentals

No issues with code that I could see. Good job with naming conventions, I know I have trouble with the C# specific ones, as Java is not to particular about the one's they use.

Generics Homework/GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,79 @@
...
+                Node<T> temp = Next;
+                while (temp.Next != this)
+                {
+                    temp = Next.Next;
+                }
+                temp.Next = new Node<T>(value);
+                // To circle back to first node
+                temp.Next.Next = this;
+            }
+        }
```
mmwoodfo, line 65: This may cause items to be placed in odd orders given the following scenerio...

Lets say I want to build the following linked list...
1 -> 2 -> 3 -> 4
┖────────┘

Now let's say I go about creating that list in the following unit test:

```C#
// Arrange

one.Insert(2);
var two = one.Next;

two.Insert(3);
 var three = two.Next;

 // Act
three.Insert(4);
var four = three.Next;

// Assert
Assert.AreEqual(one, two.Next.Next.Next);
Assert.AreEqual(two, one.Next);
Assert.AreEqual(four, three.Next);
```

Your code will output this as the list:
1 -> 3 -> 2 -> 4
┖────────┘
This is due to this portion of code when the line `two.Insert(3)` is called.
```C#
// Current state : temp.Next = 2, value = 3, temp = 1, temp.Next.Next = 1, this = 2
temp.Next = new Node<T>(value); // What you are doing here is replacing 1 -> 2 with 1 -> 3
//                                                                      ┖────┘          ┖┘
// Current state : temp.Next = 3, temp = 1, temp.Next.Next = 3, this = 2
temp.Next.Next = this; // What you are doing here is 1 -> 3 -> 3 with 1 -> 3 -> 2 (with the correct circular structure). Effectivly changing the order of your list. I know you have an insert at end method, but there shouldn't be a seperate method for this funcunality.
```

That is the first issue. The second issue is inserting into the middle of a list - there is a possibility of getting stuck in an infinite loop. Let's say this is your new list structure and this is the code to go along with it.
1 -> 2 -> 3 -> 4 -> 5
┖───────────┘

```C#
// Arrange

one.Insert(2);
var two = one.Next;

two.Insert(3);
 var three = two.Next;

three.Insert(4);
var four = three.Next;

 // Act
four.Insert(4);
var five = four.Next;

// Assert
Assert.AreEqual(one, two.Next.Next.Next.Next);
Assert.AreEqual(two, one.Next);
Assert.AreEqual(four, three.Next);
Assert.AreEqual(five, four.Next);
```

In the act portion of this test, you will hit an infinite loop. Here is why that is:

**First iteration of the loop**
Because of the ordering issue in your list, your current list looks like this: 1 -> 3 -> 2 -> 4
                                                                                                                     ┖────────┘
When executing 'four.Insert(5)' four doesn't actually refer to the four in our list above, it refers to 3 _(because of the ordering issue above)._ So what we are really doing with 'four.Insert(5)' is trying to insert a 5 between the 3 and the 2 (a valid operation).
```C#
// Current state: this = 3, temp = 2, temp.Next = 4
while (temp.Next != this)
    {
        temp = Next.Next;
    }
```
**Second iteration**
```C#
// Current state: this = 3, temp = 4, temp.Next = 1
while (temp.Next != this)
    {
        temp = Next.Next;
    }
```
**Third iteration**
```C#
// Current state: this = 3, temp = 4, temp.Next = 1
while (temp.Next != this)
    {
        temp = Next.Next;
    }
```
**Fourth iteration**
```C#
// Current state: this = 3, temp = 4, temp.Next = 1
while (temp.Next != this)
    {
        temp = Next.Next;
    }
```
So here you see the loop - there is just the right number of elements (4) to skip over the one we want to insert next to when the `temp = Next.Next` is called.

mmwoodfo:
The code review you did seems to be late. They are due Wednesday nights.

PR #61, opened 2021-02-09
Assignment5

jschofield1:
Collaborated with @coproduct

        Define the Node class

Extra Credit

    Implement Systm.Collections.Generic.ICollection<T> on the Node class ✔

Fundamentals

### DIFF
--- a/Assignment5/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/Assignment5/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -1,0 +1,94 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace GenericsHomework.Tests
+{
+    [TestClass]
+    public class NodeTests
+    {
+        [TestMethod]
+        public void Node_PassesDifferentDataTypes_NodesWithDifferentTypesAreCreated()
+        {
+            Node<int> first = new Node<int>(1);
+            Node<double> second = new Node<double>(2.0);
+            Node<string> third = new Node<string>("three");
+            Node<char> fourth = new Node<char>('4');
+
+            Assert.AreEqual<int>(1, first.Data);
+            Assert.AreEqual<double>(2.0, second.Data);
+            Assert.AreEqual<string>("three", third.Data);
+            Assert.AreEqual<char>('4', fourth.Data);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ToString_PassesNullNode_ThrowsArgumentNullException()
+        {
+            Node<string> node = new Node<string>(null!);
+
+            node.ToString();
+        }
+
+        [TestMethod]
+        public void ToString_PassesDifferentDataTypes_ReturnsValuesAsStrings()
+        {
+            int intNum = 1;
+            double dblNum = 2.0;
+            string strNum = "three";
+            char chrNum = '4';
+
+            Node<int> first = new Node<int>(intNum);
+            Node<double> second = new Node<double>(dblNum);
+            Node<string> third = new Node<string>(strNum);
+            Node<char> fourth = new Node<char>(chrNum);
+
+            Assert.AreEqual<string>(intNum.ToString(), first.ToString());
+            Assert.AreEqual<string>(dblNum.ToString(), second.ToString());
+            Assert.AreEqual<string>(strNum.ToString(), third.ToString());
+            Assert.AreEqual<string>(chrNum.ToString(), fourth.ToString());
+        }
+
+        [TestMethod]
+        public void Node_CreateTwoNodesWithDifferentValues_NodesAreNotEqual()
+        {
+            Node<int> first = new Node<int>(1);
+            Node<int> second = new Node<int>(2);
+
+            Assert.AreNotEqual<Node<int>>(first, second);
+        }
+
+        [TestMethod]
+        public void Node_ConstructNewNodeWithNext_ConstructsNextNodeAfterHead()
+        {
+            Node<int> node = new Node<int>(1);
+            Node<int> nextNode = node.Next;
+
+            Assert.AreEqual<Node<int>>(node, nextNode);
+        }
+
+        [TestMethod]
+        public void Insert_InsertNewNodeonHead_NextOnLastNodeLoopsToHead()
+        {
+            Node<int> node = new Node<int>(1);
+            
+            node.Insert(2);
+
+            Assert.AreEqual<int>(1, node.Data);
+            Assert.AreEqual<int>(2, node.Next.Data);
+            Assert.AreEqual<int>(1, node.Next.Next.Data);
+        }
+
+        [TestMethod]
+        public void Clear_ClearsArrayOnHead_OnlyHeadRemains()
+        {
+            Node<int> node = new Node<int>(1);
+            
+            node.Insert(2);
+            node.Insert(3);
+            node.Clear();
+
+            Assert.AreEqual<int>(1, node.Data);
+            Assert.AreEqual<int>(1, node.Next.Data);
+        }
+    }
+}
--- a/Assignment5/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
+++ b/Assignment5/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
@@ -7,6 +7,24 @@ namespace GenericsHomework.Tests
     public class NodeTests
     {
         [TestMethod]
+        public void GetArrayLength_CreateMultipleNodes_ArrayEqualsExpectedLength()
+        {
+            Node<int> node = new Node<int>(1);
+            
+            node.Insert(2);
+
+            Assert.AreEqual(2, node.Count);
+        }
+
+        [TestMethod]
+        public void IsReadOnly_ChecksIfNodeIsReadOnly_ReturnsFalse_()
+        {
+            Node<int> node = new Node<int>(1);
+            
+            Assert.IsFalse(node.IsReadOnly);
+        }
+
+        [TestMethod]
         public void Node_PassesDifferentDataTypes_NodesWithDifferentTypesAreCreated()
         {
             Node<int> first = new Node<int>(1);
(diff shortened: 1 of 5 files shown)


MeguminXIII:
I will review

MeguminXIII (commented):
Define the Node class
The Next property setter should be private. ❌
setter is not set to private, an example: public Property prop { get; private set; }

Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌
Did not do this option.

Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ❌
No comment in or around Clear() talking about garbage collection

Extra Credit

Implement Systm.Collections.Generic.ICollection<T> on the Node class ✔
Fundamentals
For this assignment, always use Assert.AreEquals<T>() (the generic version)❌
Did not use generic AreEquals in:
GetArrayLength_CreateMultipleNode_ArrayEqualsExpectedLength()
CopyTo_CopyValuesToArray_ArrayHoldsExpectedValues()
Remove_RemoveSpecificValue_ExpectedValueRemoved()
GetEnumerator_ForeachToIterate_AddsExpectedValues()

Good Job

jschofield1:
        Define the Node class

Extra Credit

    Implement Systm.Collections.Generic.ICollection<T> on the Node class ✔

Fundamentals

Assignment5/GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,173 @@
...
+            get
+            {
+                return _Data!;
+            }
+            private set
+            {
+                _Data = value ??
+                    throw new ArgumentNullException(nameof(value));
+            }
+        }
```
mmwoodfo, line 22: You don't need to do validation on `_Data` here. Just keep the private variable and remove the public property.

Assignment5/GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        private Node<T>? _Next;
+
+        public Node<T> Next
+        {
+            get
+            {
+                return _Next!;
+            }
+            private set
```
mmwoodfo, line 32: Submitted late in [this](https://github.com/IntelliTect-Samples/EWU-CSCD371-2021-Winter/pull/61/commits/547237a1d0eada310b237041867c824fe8289760) PR.

Assignment5/GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,173 @@
...
+            return Data.ToString()!;
+        }
+
+        public void Insert(T value)
+        {
+            Node<T> newNode = new(value);
+            Next = newNode;
+        }
+
+        /* No need to worry about garbage collection in the case of a circular list, garbage collector handles circular references */
```
mmwoodfo, line 86: Submitted late in [this](https://github.com/IntelliTect-Samples/EWU-CSCD371-2021-Winter/pull/61/commits/547237a1d0eada310b237041867c824fe8289760) PR.

Assignment5/GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,173 @@
...
+                return _Data!;
+            }
+            private set
+            {
+                _Data = value ??
+                    throw new ArgumentNullException(nameof(value));
+            }
+        }
+
+        private Node<T>? _Next;
```
mmwoodfo, line 24:
```suggestion
        private Node<T> _Next;
```
If you set this in the constructor it does not need to be nullable.

Assignment5/GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        public bool IsReadOnly
+        {
+            get;
+        }
+
+        public Node(T value)
+        {
+            Data = value;
+            Next = this;
```
mmwoodfo, line 69:
```suggestion
            _Data = value;
            _Next = this;
```

Assignment5/GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        public override string ToString()
+        {
+            if (Data is null) 
+                throw new ArgumentNullException(nameof(Data));
```
mmwoodfo, line 75:
```suggestion
            if (Data is null) 
                throw new ArgumentNullException(nameof(Data));
```
Consider returning an empty string if `Data` is null.

Assignment5/GenericsHomework/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            return Data.ToString()!;
```
mmwoodfo, line 77:
```suggestion
            return Data?.ToString() ?? "";
```
This will return an empty string when null. Avoid using `!` if you can help it.

Assignment5/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            Node<int> first = new Node<int>(intNum);
+            Node<double> second = new Node<double>(dblNum);
+            Node<string> third = new Node<string>(strNum);
+            Node<char> fourth = new Node<char>(chrNum);
+
+            Assert.AreEqual<string>(intNum.ToString(), first.ToString());
+            Assert.AreEqual<string>(dblNum.ToString(), second.ToString());
+            Assert.AreEqual<string>(strNum.ToString(), third.ToString());
+            Assert.AreEqual<string>(chrNum.ToString(), fourth.ToString());
```
mmwoodfo, line 84:
```suggestion
            Assert.AreEqual(intNum.ToString(), first.ToString());
            Assert.AreEqual(dblNum.ToString(), second.ToString());
            Assert.AreEqual(strNum.ToString(), third.ToString());
            Assert.AreEqual(chrNum.ToString(), fourth.ToString());
```

Assignment5/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        [TestMethod]
+        public void Insert_InsertNewNodeOnHead_NextOnLastNodeLoopsToHead()
+        {
+            Node<int> node = new Node<int>(1);
+            
+            node.Insert(2);
+
+            Assert.AreEqual<int>(1, node.Data);
+            Assert.AreEqual<int>(2, node.Next.Data);
+            Assert.AreEqual<int>(1, node.Next.Next.Data);
```
mmwoodfo, line 96: You should test more than two nodes being inserted into the list. You should also test the order in which the nodes get added to the list. When writing unit tests the goal is to write unit tests that stress test your code. Hit the limits, test edge cases, and hopefully reveal bugs.

The following unit test will cause a failure in your code.
```C#
// 1 -> 2 -> 3
// ┖────┘

// Arrange

// Act
one.Next.Insert(2);
var two = one.Next;

two.Next.Insert(3);
var three = two.Next;

```

```
1 -> 2 -> 3
┖─────────┘
```

You'll have this:
```
1 -> 3
┖────┘

2 -> 1 -> 3
     ┖────┘
```

There are easy ways within visual studio to look at your variables and data structures. [Here are some tips for that](https://devblogs.microsoft.com/devops/7-ways-to-look-at-the-values-of-variables-while-debugging-in-visual-studio/).

Assignment5/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            int[] copiedArray = new int[3];
+            int[] expectedArray = { 1, 2, 3 };
+
+            Node<int> node = new Node<int>(1);
+
+            node.Insert(2);
+            node.Insert(3);
+            node.CopyTo(copiedArray, 0);
+
+            Assert.AreEqual<string>($"{expectedArray.ToString()}", $"{copiedArray.ToString()}");
```
mmwoodfo, line 132: This is not doing what you think it's doing. The output of this test is: `Assert.AreEqual("System.Int32[]", "System.Int32[]");`. Which is true. Hence why your test passes.

You should use `Assert.IsTrue(expectedArray.SequenceEqual(copiedArray));` instead. This will fail, because of the comment on line 96 above.

Assignment5/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            
+            node.Insert(2);
+
+            Assert.AreEqual<int>(1, node.Data);
+            Assert.AreEqual<int>(2, node.Next.Data);
+            Assert.AreEqual<int>(1, node.Next.Next.Data);
+        }
+
+        [TestMethod]
+        public void Clear_ClearsArrayOnHead_OnlyHeadRemains()
```
mmwoodfo, line 100: It would be good to have a unit test that tests if the garbage collector ever cleans up the detached nodes.

Assignment5/GenericsHomework/GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            Assert.AreEqual<Node<int>>(node, nextNode);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ToString_PassesNullNode_ThrowsArgumentNullException()
+        {
+            Node<string> node = new Node<string>(null!);
+
+            node.ToString();
```
mmwoodfo, line 65: This is not the line throwing the exception. This unit test is misleading in what it's actually testing. Your exception gets thrown on line 19 of your `Node` class when you try to set a null value to your `Data` property.

mmwoodfo:
@jschofield1 I can't find a code review from you for Assignment5.

jschofield1:
> 
> 
> @jschofield1 I can't find a code review from you for Assignment5.

We were told on the first day of class that if we paired with a classmate we were not required to do peer reviews. My review of my own code is on the initial comment.

mmwoodfo:
> > @jschofield1 I can't find a code review from you for Assignment5.
> 
> We were told on the first day of class that if we paired with a classmate we were not required to do peer reviews. My review of my own code is on the initial comment.

@jschofield1 No problem. I missed that you mentioned in your comment you paired. I'll update both your score and @coproduct's score accordingly. Note that this may take a few days. Thank you!

PR #63, opened 2021-02-23
Assignment6

worseTyler:


Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,59 @@
...
         // 4.
-        public IEnumerable<IPerson> People => throw new NotImplementedException();
+        public IEnumerable<IPerson> People => CsvRows.Select(item =>
+        {
+            string[] items = item.Split(',');
+            Person person = new(items[1], items[2], new Address(items[4], items[5], items[6], items[7]), items[3]);
+            return person;
+        }).OrderBy(item => item.Address.State)
+            .ThenBy(item => item.Address.City)
+            .ThenBy(item => item.Address.Zip);
```
mmwoodfo, line 44: You should first select, then order, then select again. Also consider using an enum.
```C#
        public IEnumerable<IPerson> People =>
            CsvRows
                .Select(row => row.Split(","))
                .OrderBy(line => line[6]) // State
                .ThenBy(line => line[5]) // City
                .ThenBy(line => line[7]) // Zip
                .Select(item => new Person(
                    ...
                 ));
```

Assignment/Assignment/SampleData.cs
```diff
@@ -1,30 +1,59 @@
...
 
         // 6.
         public string GetAggregateListOfStatesGivenPeopleCollection(
-            IEnumerable<IPerson> people) => throw new NotImplementedException();
+            IEnumerable<IPerson> people) => people.Select(item => item.Address.State).Distinct().Aggregate(string.Empty, (item, next) =>
+            {
+                item += next;
+                item += ", ";
+                return item;
+            }).TrimEnd(',', ' ');
```
mmwoodfo, line 57: Consider using string interpolation here. There is a better way to do this that doesn't require `TrimEnd`

Assignment/Assignment/Node.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        public IEnumerable<T> ChildItems(int maximum)
+        {
+			return this.AsEnumerable<T>().Take(maximum);
```
mmwoodfo, line 101: 👍🏻

Assignment/Assignment/Node.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            Node<T> traverse = this;
+            do
+            {
+                yield return traverse.Value;
+                traverse = traverse.Next;
+            } while (traverse != this);
```
mmwoodfo, line 91: 👍🏻

PR #69, opened 2021-03-04
Assignment7

alexvthornton:
partner: @kaykay38


alexvthornton:
In your tests use https://google.com repeatedly for the urls parameter.
Implement an async (using TPL) method, DownloadTextRepeatedlyAsync(int repetitions, params string[] urls), that uses a parallel loop to call DownloadTextAsync repeatedly for repetitions times and returns the total number of characters downloaded across all iterations. ✔
Write a test to show that the code is not synchronized as is. ✔
Synchronize and test using lock. ✔
Synchronize and test using System.Threading.Interlocked. ✔
Demonstrate that synchronization is not required when using System.Threading.ThreadLocal<T>. ✔
Testing

Configure all tests to run in parallel. ✔
Extra Credit

Submit a PR with unit tests for at least 3 code listings in Chapters 20-23 (see https://github.com/IntelliTect/EssentialCSharp/tree/v8.0/src) where the the listing currently has no unit tests. ❌
Fundamentals

Choose simplicity over complexity ✔

Lectures/2020.01.07-UnitTesting/Assignment7.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.Linq;
+using System.Threading;
+
+
+namespace Assignment7
+{
+
+    static class methods
```
breaman, line 13: Classes should be Pascal Case and match the filename

breaman (commented):
Overall really good work on the assignment. Only had one issue around class names, but nothing too major.

PR #70, opened 2021-03-04
Assignment7

MeguminXIII:


cmartinezaraujo:
I will review.

Assignment7/Assignment7/Assignment7Tests.cs
```diff
@@ -0,0 +1,78 @@
...
+        [ExpectedException(typeof(AggregateException))]
+        public void DownloadTextRepeatedlyAsync_NullProgress_ExceptionExpected()
+        {
+            CancellationTokenSource source = new();
+            CancellationToken token = source.Token;
+            Assert.AreEqual<int>(0, Assignment7.DownloadTextRepeatedlyAsync(100, token, null!, "https://google.com").Result);
+        }
+
+        [TestMethod]
+        public void DownloadTextRepeatedlyAsync_Cancel_GetsCancelled()
```
cmartinezaraujo, line 65: You might want to check on this test. It seems to pass but only because the result seems to be under 10000000 for the 100 repetitions letting it run all the way through. Expected:<4831989>. Actual:<4829956> these are the results from an expected run of 100 repetitions with no call to cancel vs the one in the test that is supposed to get cancelled. Looks like the if statement just never evaluates to true with the current values  in the test because taking it off gets it to cancel.

cmartinezaraujo (commented):

In your tests use https://google.com repeatedly for the urls parameter.
Implement an async (using TPL) method, DownloadTextRepeatedlyAsync(int repetitions, params string[] urls), that uses a parallel loop to call DownloadTextAsync repeatedly for repetitions times and returns the total number of characters downloaded across all iterations. ✔

Write a test to show that the code is not synchronized as is. ✔

Synchronize and test using lock. ✔
Synchronize and test using System.Threading.Interlocked. ✔
Demonstrate that synchronization is not required when using System.Threading.ThreadLocal<T>. ✔

Testing
Configure all tests to run in parallel. ❌

Extra Credit
Submit a PR with unit tests for at least 3 code listings in Chapters 20-23 (see https://github.com/IntelliTect/EssentialCSharp/tree/v8.0/src) where the the listing currently has no unit tests. ❌✔

Fundamentals
Choose simplicity over than complexity ✔

Looks like the only things you might want to look at are the cancellation test and running the tests in parallel. I had also just clicked on the parallel option in the test explorer but I had to manually do it for your project as well so it looks like it wont automatically trigger that option when someone else opens your project.

I assume you already saw the links Kevin showed us at the end of class the other day but just in case 

https://docs.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2019

https://docs.microsoft.com/en-us/azure/devops/pipelines/test/parallel-testing-vstest?view=azure-devops#:~:text=Visual%20Studio%20Test%20Platform%20can%20run%20test%20assemblies,handing%20it%20tests%20in%20an%20assembly%20to%20execute.

Other than that everything else looks great!

Assignment7/UnitTest.runsettings
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+	<MaxCpuCount>1</MaxCpuCount>
```
breaman, line 3: This enables parallel running, but since you used a 1 instead of a 0, this really isn't letting things run in parallel since it is only going to use one CPU.

breaman (commented):
Nice work on the assignment. I prefer to try and avoid having a class name the same as my namespace, but didn't dock for that just wanted to point out that sometimes it can cause issues

PR #71, opened 2021-03-04
Assignment7

jschofield1:
Collaborated with @coproduct


jschofield1:
Place all code in a single unit testing project called Assignment7.
1.	In this example, we are intentionally System.Net.WebClient.DownloadString() to simulate a high-latency operation - pretending there is no async. version.
•	Implement an async (using TPL) method, DownloadTextAsync(params string[] urls), that calls System.Net.WebClient.DownloadString() and returns the total number of characters downloaded from all the urls. ✔ 
o	You do not need to download the urls in parallel.
o	In your tests use https://google.com repeatedly for the urls parameter.
•	Implement an async (using TPL) method, DownloadTextRepeatedlyAsync(int repetitions, params string[] urls), that uses a parallel loop to call DownloadTextAsync repeatedly for repetitions times and returns the total number of characters downloaded across all iterations. ✔
•	Add support to DownloadTextRepeatedlyAsync for cancellation so that the method can be cancelled before all the repetitions occur. ✔
•	Add support to DownloadTextRepeatedlyAsync for observing the progress (IProgress<T>) that shows the percentage of repetitions that have completed. ✔
•	Be sure to add any synchronization if needed. ✔
1.	Turn in the solutions discussed/solved in class using the following un-synchronized code:
•	Write a test to show that the code is not synchronized as is. ✔
•	Synchronize and test using lock. ✔
•	Synchronize and test using System.Threading.Interlocked. ✔
•	Demonstrate that synchronization is not required when using System.Threading.ThreadLocal<T>. ✔
Testing
•	Configure all tests to run in parallel. ✔
Extra Credit
•	Submit a PR with unit tests for at least 3 code listings in Chapters 20-23 (see https://github.com/IntelliTect/EssentialCSharp/tree/v8.0/src) where the the listing currently has no unit tests. ❌

Assignment7/Assignment7/Assignment7/test.runsettings.xml
```diff
@@ -0,0 +1,109 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+	<!-- Configurations that affect the Test Framework -->
+	<RunConfiguration>
+		<MaxCpuCount>1</MaxCpuCount>
```
breaman, line 5: While this enables parallel running, leaving this value as a 1 really then says "you can run in parallel, but you can only use one CPU" which really doesn't allow for parallel running. Should have been a 0 or a number greater than 1.

breaman (commented):
Nice work on the assignment overall. Did notice that you enabled parallel test runs, but then restricted it to only 1 CPU, didn't dock, but just wanted to point that out.

PR #73, opened 2021-03-04
Assignment7

cmartinezaraujo:


worseTyler:
I will review this

worseTyler:
Instructions
DUE: All homework is allowed an additional two days (so rather than submitting for peer review Monday it may be posted Wednesday). Final code should be submitted by midnight on Saturday, March 6.
Place all code in a single unit testing project called Assignment7.
1.	In this example, we are intentionally System.Net.WebClient.DownloadString() to simulate a high-latency operation - pretending there is no async. version.
•	Implement an async (using TPL) method, DownloadTextAsync(params string[] urls), that calls System.Net.WebClient.DownloadString() and returns the total number of characters downloaded from all the urls. ✔
o	You do not need to download the urls in parallel.
o	In your tests use https://google.com repeatedly for the urls parameter.
•	Implement an async (using TPL) method, DownloadTextRepeatedlyAsync(int repetitions, params string[] urls), that uses a parallel loop to call DownloadTextAsync repeatedly for repetitions times and returns the total number of characters downloaded across all iterations. ❌
•	Add support to DownloadTextRepeatedlyAsync for cancellation so that the method can be cancelled before all the repetitions occur. ✔
•	Add support to DownloadTextRepeatedlyAsync for observing the progress (IProgress<T>) that shows the percentage of repetitions that have completed. ✔
•	Be sure to add any synchronization if needed. ✔

(I couldn't find anything for this section below)
2.	Turn in the solutions discussed/solved in class using the following un-synchronized code:
•	Write a test to show that the code is not synchronized as is. ❌
•	Synchronize and test using lock. ❌
•	Synchronize and test using System.Threading.Interlocked. ❌
•	Demonstrate that synchronization is not required when using System.Threading.ThreadLocal<T>. ❌
Testing
•	Configure all tests to run in parallel. ❌
Extra Credit
•	Submit a PR with unit tests for at least 3 code listings in Chapters 20-23 (see https://github.com/IntelliTect/EssentialCSharp/tree/v8.0/src) where the the listing currently has no unit tests. ❌
Fundamentals
•	Ensure you enable nullable reference types, net5 targeted, C# 9.0, warnings as errors, and enabled .NET analyzers for both projects ✔
•	Fully unit test the assignment ✔
(It might be good to add some unit tests for edge cases such as passing in a null params because your methods wouldn’t know how to handle that)
•	Choose simplicity over than complexity ✔

Assignment7/Assignment7.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        public static async Task<int> DownloadTextRepeatedlyAsynch(IProgress<int> progress, CancellationToken cancellationToken, int repetitions, params string[] urls)
+        {
+
+            return await Task.Run(async () =>
+            {
+                //int progress = 0;
+                int total = 0;
+                for (int i = 0; i < repetitions && !(cancellationToken.IsCancellationRequested); i++)
```
worseTyler, line 28: Not a parallel loop

Assignment7/Assignment7.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Assignment7
+{
+    public class Assignment7
+    {
+        public static async Task<int> DownloadTextAsync(params string[] urls)
+        {
+
+            WebClient client = new WebClient();
+
+            return await Task.Run(() => urls.Aggregate(0, (total, next) => total + client.DownloadString(next).Length));
```
worseTyler, line 18: I like how you were able to make this one line, it looks clean

Assignment7/Assignment7.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Assignment7
+{
+    public class Assignment7
+    {
+        public static async Task<int> DownloadTextAsync(params string[] urls)
```
worseTyler, line 13: What would happen if somehow a null string array was entered (null!)

Assignment7/Assignment7Tests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Assignment7
+{
+    [TestClass]
+    public class Assignment7Tests
+    {
+        string[] urls = { "https://google.com", "https://google.com", "https://google.com", "https://google.com", "https://google.com", "https://google.com" };
+        WebClient client = new WebClient();
+
+        public int repetitionsCall(int repetitions)
```
worseTyler, line 15: Bad naming convention, needs to be PascalCased (also the name is a little vague, since it only used in testing code you can give it a really specific name which might be beneficial)

Assignment7/Assignment7Tests.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            Assert.IsTrue(Assignment7.DownloadTextAsync("https://google.com").Result > 0);
+        }
+
+        [TestMethod]
+        public void DownloadTextAsyncCounts_ReturnsAccurateCountWithin1000Characters()
+        {
+            int characterExpected = urls.Aggregate(0, (total, next) => total + client.DownloadString(next).Length);
+
+            Assert.IsTrue((Assignment7.DownloadTextAsync(urls).Result-characterExpected) < 1000);
```
worseTyler, line 35: I like this idea for a test, I didn't think about something like this myself and it seems like a pretty good idea. However, I don't know if google ever has instances where more than a thousand characters are added to the page at a time which might break this test from time to time

Assignment7/Assignment7Tests.cs
```diff
@@ -0,0 +1,97 @@
...
+        [TestMethod]
+        public void DownloadTextRepeatedlyAsynch_Returns0WhenCanceledBeforCall()
+        {
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+            CancellationToken cancellationToken = cancellationTokenSource.Token;
+            cancellationTokenSource.Cancel();
+            var progress = new Progress<int>(progressPercent => Console.WriteLine(progressPercent));
+
+
+            int result = (Assignment7.DownloadTextRepeatedlyAsynch(progress, cancellationToken, 100, urls)).Result;
```
worseTyler, line 62: Cancels task before even calling function, good idea for a test

Assignment7/Assignment7.cs
breaman: As @worseTyler  pointed out, you should have put in a null check since you will throw a very obscure exception (Object reference not set to an instance of an object) if urls happens to be null.

breaman (commented):
The download code was written nice and clean, but looks like you missed a chunk of the assignment around testing the unsynchornized code from the book.

PR #74, opened 2021-03-05
Assignment7

Sithale:
Methods work and there are tests for both methods.
The last two files were a failed attempt of opening and creating a separate project for testing and seeing how some other tests were done, they were deleted shortly after being made. For some reason the commit registered the old files that no longer existed.
Files 1 - 4 are the final result so far.


MeguminXIII:
I will review

Assignment7/Assignment7/ThreadingMethod.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public static async Task<int> DownloadTextRepeatedlyAsync(int rep, CancellationToken cancel, IProgress<double> prog, params string[] urls)
+        {
+            if (rep < 0) throw new AggregateException(nameof(rep));
+            if (prog is null) throw new AggregateException(nameof(prog));
+
+            int count = 0;
+            int x = 0;
+
+            for (int y = 0; y < rep && !cancel.IsCancellationRequested; y ++)
```
MeguminXIII, line 34: I think we are supposed to use something like a parallel.for loop. I looked up some documentation if you want to use it, not sure what other kind of parallel loops there are: https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/how-to-write-a-simple-parallel-for-loop

MeguminXIII (commented):
In your tests use https://google.com repeatedly for the urls parameter.✔
Implement an async (using TPL) method, DownloadTextRepeatedlyAsync(int repetitions, params string[] urls), that uses a parallel loop to call DownloadTextAsync repeatedly for repetitions times and returns the total number of characters downloaded across all iterations. ❌
I think we are supposed to use something like a parallel.for loop

Write a test to show that the code is not synchronized as is. ❌
Did not use the code we looked at in class, 
Code can be found here: https://raw.githubusercontent.com/IntelliTect/EssentialCSharp/v8.0/src/Chapter22/Listing22.01.UnsychronizedState.cs

Synchronize and test using lock. ❌
Synchronize and test using System.Threading.Interlocked. ❌
Demonstrate that synchronization is not required when using System.Threading.ThreadLocal<T>. ❌
Did not do these

Testing
Configure all tests to run in parallel. ❌
I think we are supposed to use a .runsettings file

Extra Credit
Submit a PR with unit tests for at least 3 code listings in Chapters 20-23 (see https://github.com/IntelliTect/EssentialCSharp/tree/v8.0/src) where the the listing currently has no unit tests. ❌
Did not do

Fundamentals
Choose simplicity over than complexity ✔

Code looks good, just missing a few pieces of the assignment and need to change a loop into a parallel loop

Assignment7/Assignment7/ThreadingMethod.cs
breaman: @MeguminXIII  was right, this is not going to do anything in parallel, it will wait for each call to DownloadTextAsync to complete before making the next call.

breaman (commented):
Looks like about half the assignment was completed. You were missing the testing and modifying of the un-synchronized code as well as didn't make the DownloadTextRepeatedlyAsync use a parallel for. I also didn't see where you enabled the unit tests to run in parallel.

PR #76, opened 2021-03-09
Assignment8

MeguminXIII:


MeguminXIII:
Done with assignment and did all extra credit

Assignment8/Assignmnet8/Assignment8.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <OutputType>WinExe</OutputType>
+    <TargetFramework>net5.0-windows</TargetFramework>
+    <UseWPF>true</UseWPF>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	<Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MaterialDesignThemes" Version="4.0.0" />
+    <PackageReference Include="ShowMeTheXAML.MSBuild" Version="2.0.0" />
```
Keboo, line 13: Though this is used in the Material Design Demo app, it is not something that is needed when consuming the library.

Assignment8/Assignmnet8/Assignment8.csproj
```diff
@@ -0,0 +1,22 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <Reference Include="System.Windows.Forms.VisualStudio.15.0">
```
Keboo, line 17: Is there a reason for including this reference? Typically you do not need System.Windows.Forms libraries within WPF

Assignment8/Assignmnet8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,111 @@
...
+namespace Assignmnet8
+{
+    /// <summary>
+    /// Interaction logic for MainWindow.xaml
+    /// </summary>
+    public partial class MainWindow : Window
+    {
+        private readonly DispatcherTimer _DispatcherTimer = new();
+        private DateTime _TimeAtStart = DateTime.Now;
+        private int Loop = 1;
```
Keboo, line 21:
```suggestion
        private int _Loop = 1;

```

Assignment8/Assignmnet8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,111 @@
...
+        private string _Details;
+        private bool _StartStop = true;
+
+        public MainWindow()
+        {
+            InitializeComponent();
+
+            _DispatcherTimer.Tick += new EventHandler(DispatcherTimer_Tick);
+            _DispatcherTimer.Interval = TimeSpan.FromSeconds(1);
```
Keboo, line 30: This works, but you will notice occasionally the timer skips displaying a given second. The reason is because the interval is an approximate. Typically you will set your interval to roughly half of the accuracy you want.

Assignment8/Assignmnet8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,111 @@
...
+        private DateTime _TimeAtStart = DateTime.Now;
+        private int Loop = 1;
+        private string _Details;
+        private bool _StartStop = true;
+
+        public MainWindow()
+        {
+            InitializeComponent();
+
+            _DispatcherTimer.Tick += new EventHandler(DispatcherTimer_Tick);
```
Keboo, line 29: Event handler can be simplified.
```suggestion
            _DispatcherTimer.Tick += DispatcherTimer_Tick;

```

Assignment8/Assignmnet8/MainWindow.xaml.cs
```diff
@@ -0,0 +1,111 @@
...
+{
+    /// <summary>
+    /// Interaction logic for MainWindow.xaml
+    /// </summary>
+    public partial class MainWindow : Window
+    {
+        private readonly DispatcherTimer _DispatcherTimer = new();
+        private DateTime _TimeAtStart = DateTime.Now;
+        private int Loop = 1;
+        private string _Details;
```
Keboo, line 22: This is declared as non-null but is not set in the constructor, so it will be null

Keboo (commented):
# Assignment
- Window content resize appropriately: ✔
- Can start and stop a timer: ✔
- Past timers are displayed in a list with duration and description: ✔
- Past timers can be deleted by pressing the delete key: ✔

## Extra Credit
- App uses theming library: ✔
- App allows for savings and loading time entries: ✔

PR #79, opened 2021-03-09
Assignment 8 finished. Looks ugly but works.

cmartinezaraujo:
Assignment 8

### DIFF
--- a/TimeTracker/MainWindow.xaml
+++ b/TimeTracker/MainWindow.xaml
@@ -1,0 +1,70 @@
+﻿<Window x:Class="TimeTracker.MainWindow"
+        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+        xmlns:local="clr-namespace:TimeTracker"
+        mc:Ignorable="d"
+        Title="MainWindow" Height="450" Width="800">
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <!-- Default Height is 1* -->
+            <RowDefinition />
+        </Grid.RowDefinitions>
+
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition />
+            <ColumnDefinition />
+        </Grid.ColumnDefinitions>
+        <StackPanel x:Name = "LeftPanel">
+            <Button 
+                Content="Start Timer"
+                Padding="10,20,10,20"
+                Grid.Row="0"
+                Grid.Column="0"
+                x:Name="TimerStartButton" 
+                Click="TimerStartButton_Click"/>
+            
+            <Button 
+                Padding="10,20,10,20"
+                Content="Stop Timer" 
+                Grid.Row="1"
+                Grid.Column="0"
+                x:Name="TimerStopButton" 
+                Click="TimerStopButton_Click"/>
+
+            <Button 
+                Padding="10,20,10,20"
+                Content="Reset Timer" 
+                Grid.Row="1"
+                Grid.Column="0"
+                x:Name="TimerReset" Click="TimerReset_Click"/>
+
+            <Button 
+                Padding="10,20,10,20"
+                Content="Log Timer" 
+                Grid.Row="1"
+                Grid.Column="0"
+                x:Name="LogTimer" Click="LogTimer_Click"/>
+            
+
+            <TextBlock Text="Timer Description" />
+            <TextBox 
+                 Grid.Column="0"
+                 Padding="10,30,10,30"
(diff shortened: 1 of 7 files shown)


MeguminXIII:
I will review

TimeTracker/MainWindow.xaml
```diff
@@ -0,0 +1,70 @@
...
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+        xmlns:local="clr-namespace:TimeTracker"
+        mc:Ignorable="d"
+        Title="MainWindow" Height="450" Width="800">
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <!-- Default Height is 1* -->
```
MeguminXIII, line 12: Not sure if this comment adds anything? It might be fine though

TimeTracker/MainWindow.xaml.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// Interaction logic for MainWindow.xaml
+    /// </summary>
+
+    public partial class MainWindow : Window
+    {
+        private DateTime startTime;
+        private DateTime endTime;
+        private TimeSpan timeDifference;
+        private bool isReset;
+        private string timerDescription;
```
MeguminXIII, line 20: Is not in PascalCase which is used for class wide fields, this is camelCase which is used in methods.

MeguminXIII (commented):
The main windows should scale up and down nicely. That is the controls on the windows should expand and shink in a reasonable manner as the window size expands and shinks.❌
Timer log, and buttons vanish when window shrinks.

There should be UI elements to handle the following interactions:
Start a new timer✔
When a timer is started the UI should update, showing how much time has elapsed✔
Stop the currently running timer✔
Set a description to be associated with the current timer✔

Maintain a list of past timers that shows the description and the durration✔
I put yes on this, however it does not seem to log it correctly, ex. When the time on the timer is on 2.854 seconds it shows 2.880 on the log. Not a large difference, so it may not matter much. The difference doesn't seem to get larger than 0.021 seconds even when I let it run for a few minutes. I can't figure out why from looking at your code so it may be something on my end

Past timers should be able to be removed by pressing the delete key✔
Extra Credit

Give your app a little bit of flair with a theming library. Two popular options are Material Design in XAML and MahApps.❌
Add functionality to save and load the time entries.❌
Didn't do them

Looks good, Just need to fix the window scaling on your timer, log, and it looks like your bottom two buttons are the only ones that get cut for some reason. Maybe look at the timer logging button as well? but again that may be lag or something on my end.

MeguminXIII:
WPF Project created with nullability, and analyzers enabled: ❌
Sorry I missed this somehow, But you need to add in nullability and enable analyzers

TimeTracker/MainWindow.xaml.cs
```diff
@@ -0,0 +1,102 @@
...
+            _IsReset = true;
+
+        }
+
+        private void PastTimersLog_KeyDown(object sender, KeyEventArgs e)
+        {
+            e.Handled = true;
+            if (e.Key == Key.Delete)
+            {
+                PastTimersLog.Items.Remove(PastTimersLog.Items.OfType<object>().Last());
```
Keboo, line 98: This works, but is a bit unusual if you select the first item and press delete. Normal interaction would be to remove the selected item.

TimeTracker/MainWindow.xaml.cs
```diff
@@ -0,0 +1,102 @@
...
+        private void TextBox_TextChanged(object sender, System.Windows.Controls.TextChangedEventArgs e)
+        {
+            _TimerDescription = TimerDescriptionTextBox.Text;
+        }
+
+        private void LogTimer_Click(object sender, RoutedEventArgs e)
+        {
+            if(DispatcherTimer.IsEnabled)
+            {
+                PastTimersLog.Items.Add($"{_TimerDescription} {Convert.ToString(DateTime.Now - _StartTime)}");
```
Keboo, line 75: When inside of an interpolated string you do not need to use Convert.ToString. Parameters are automatically converted to a string.
```suggestion
                PastTimersLog.Items.Add($"{_TimerDescription} {(DateTime.Now - _StartTime)");

```

TimeTracker/MainWindow.xaml.cs
```diff
@@ -0,0 +1,102 @@
...
+        public MainWindow()
+        {
+            InitializeComponent();
+
+            DispatcherTimer = new();
+            DispatcherTimer.Interval = new TimeSpan(1);
+            DispatcherTimer.Tick += DispatcherTimer_Tick;
+        }
+
+        private void DispatcherTimer_Tick(object sender, EventArgs e)
```
Keboo, line 32: The nullable signature of the event handler expects the `sender` to be declared nullable.
```suggestion
        private void DispatcherTimer_Tick(object? sender, EventArgs e)

```

Keboo (commented):
# Assignment
- Window content resize appropriately: ✔
- Can start and stop a timer: ✔
- Past timers are displayed in a list with duration and description: ✔
- Past timers can be deleted by pressing the delete key: ✔

## Extra Credit
- App uses theming library: ❌
- App allows for savings and loading time entries: ❌

PR #81, opened 2021-03-16
Assignment9

cmartinezaraujo:
Still figuring some stuff out but application should be functional.


MeguminXIII:
I will Review

ContactApp/MainWindowViewModel.cs
```diff
@@ -0,0 +1,123 @@
...
+                EmailAddress = "Mister@gmail.com",
+                TwitterHandle = "@Mistery",
+                LastModified = DateTime.Now
+            });
+
+        }
+
+        private void OnNewContact()
+        {
+            ContactViewModel NewContact = new ContactViewModel() { FirstName = "Jane", LastName = "Doe", LastModified = DateTime.Now };
```
MeguminXIII, line 71: newContact
camelCase for method-wide, PascalCase for class wide

MeguminXIII (commented):
Assignment
WPF Project created with nullability, and analyzers enabled: ✔
App uses MVVM pattern (MainWindowViewModel): ✔
All view models are tested: ❌
Could use more tests, Testing the Save, Load, and Delete commands for example. Testing the ContactViewModel as well.

App shows a list of contacts: ✔
Can create a new contact: ✔
Can delete a contact: ✔
Can view a contact's details: ✔
Can edit a contact's details: ✔
Contact fields that are not specified are not shown while viewing using triggers: ✔
Contact contains all required fields: ✔

it says in the assignment details to sort the contact list,
"The main window should display a list of contacts, the list should be sorted by the contacts name." it doesn't say anything about it in the assignment.md though.

Extra Credit
In addition to the default contact fields, allow for adding an arbitrary set of additional fields: ❌ App allows for savings and loading contacts using Entity Framework Core (simply use SQLite): ❌
Doesn't look like it was implemented

Looks good. I think it needs more tests, and again the assignment details found https://github.com/IntelliTect-Samples/EWU-CSCD371-2021-Winter/tree/Assignment9 
says the contacts list needs to be sorted. I'm not sure why it wasn't in the Assignment.md file

ContactApp/ContactViewModel.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            get => _LastModified;
+            set
+            {
+                _LastModified = value;
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(LastModified)));
+            }
+        }
+
+        private void SetProperty(ref string field, string newValue, [CallerMemberName] string propertyName = "")
```
Keboo, line 59: This is not quite correct. The correct implementation would be the one that you have in the MainWindowViewModel.

ContactApp/ContactViewModel.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        private DateTime? _LastModified;
+
+        public DateTime? LastModified
+        {
+            get => _LastModified;
+            set
+            {
+                _LastModified = value;
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(LastModified)));
```
Keboo, line 55: Normally you would still use your SetProperty method. You only want to raise the `PropertyChanged` event if the property has actually changed. For example this raises would raise the event twice when it only needs to occur the first time:
```
DateTime now = DateTime.Now;
LastModified = now;
LastModified = now;
```

ContactApp/ContactViewModel.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            get => _EmailAddress; 
+            set => SetProperty(ref _EmailAddress!, value!);
+        }
+
+        private string? _TwitterHandle;
+        public string? TwitterHandle 
+        {
+            get => _TwitterHandle; 
+            set => SetProperty(ref _TwitterHandle!, value!);
```
Keboo, line 44: Rather than using the null forgiveness operator everywhere, it would be better to declare the parameters to SetProperty as nullable.

Keboo (commented):
# Assignment
- Can create a new contact: ✔
- Can delete a contact: ✔
- Can view a contact's deails: ✔

## Extra Credit

ContactApp/MainWindowViewModel.cs
Keboo: Agreed, casing should be camelCase for local variables.

PR #82, opened 2021-03-18
Assignment9

MeguminXIII:


cmartinezaraujo:
I will review

Assignment9/Assignment9/MainWindow.xaml
```diff
@@ -0,0 +1,194 @@
...
+            <StackPanel>
+                <TextBlock Text="First_Name" 
+                           FontWeight="ExtraBold"
+                           FontSize="20" 
+                           Visibility="{Binding FirstName, Converter={StaticResource EmptyStringVisibilityConverter}}"
+                           />
+                
+                <TextBlock Text="{Binding FirstName}" 
+                           FontSize="15" 
+                           Visibility="{Binding FirstName, Converter={StaticResource EmptyStringVisibilityConverter}}"
```
cmartinezaraujo, line 36: I think he wants us to specifically use style triggers for these according to the main assignment page.

cmartinezaraujo (commented):
Assignment
- Can create a new contact: ✔
- Can delete a contact: ✔

- Contact fields that are not specified are not shown while viewing using triggers: ✔ I am not sure if we can use the converters for the visibility on these or not  since it says "style triggers" on the main assignment page.

Extra Credit

Assignment9/Assignment9.Tests/RelayCommandTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        public void CanExecuteDelegate_CanExecuteReturnsTrueWhenAbleToExecute()
+        {
+            RelayCommand relay = new(() => System.Console.WriteLine("Hello World"), () => true);
+            Assert.IsTrue(relay.CanExecute(null));
+        }
+
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void RelayCommand_PassInNullValues_ThrowNullArgumentException()
```
Keboo, line 30: This technically doesn't test the exception thrown from the `canExecuteDelegate` since the code will always just throw at the first null.

Assignment9/Assignment9/ContactViewModel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        private DateTime _LastModifiedTime;
+
+        public DateTime LastModifiedTime
+        {
+            get => _LastModifiedTime;
+            set
+            {
+                _LastModifiedTime = value;
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(LastModifiedTime)));
```
Keboo, line 28: Typically you would want a `SetProperty` method like what you have in your `MainWindowViewModel`. That way the PropertyChanged event would only get raised when the value actually changes.

Assignment9/Assignment9/ContactViewModel.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    public class ContactViewModel
+    {
+        public string? FirstName { get; set; }
+        public string? LastName { get; set; }
+        public string? PhoneNumber { get; set; }
+        public string? EmailAddress { get; set; }
+        public string? TwitterName { get; set; }
+
+        public event PropertyChangedEventHandler? PropertyChanged;
```
Keboo, line 18: There is one small bug here. When you edit an item, the UI doesn't show the updated LastModifiedTime until after you click off of the selected time and then back on it. The reason is because this class does not extend the `INotifyPropertyChanged` interface. Without this, the Bindings are not registered to the PropertyChanged event and don't automatically update the UI (even though the event is raised)

Assignment9/Assignment9.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,85 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace Assignment9.Tests
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+        [TestMethod]
+        public void CreateNewContact_GeneratesNewContact()
```
Keboo, line 10: This in ok, but it would be better to also assert on the values being set on the new view model as well.

Keboo (commented):
# Assignment
- Can create a new contact: ✔
- Can delete a contact: ✔
- Can edit a contact's details: ✔/❌ - This mostly works, the modified time is not properly getting bound due to the missing INotifyPropertyChanged interface.

## Extra Credit

PR #83, opened 2021-03-19
Assignment9

worseTyler:


Assignment9/WpfApp/WpfApp/EmptyStringConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (!string.IsNullOrWhiteSpace((string)value))
+            {
+                return Visibility.Visible;
+            }
+            return Visibility.Collapsed;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
```
Keboo, line 23: This is a bit of an un-usual implementation. In general Convert/ConvertBack should be inverses of each other. However, in cases like this, converters sometimes are only useful in one direction. Simply leaving ConvertBack as throwing a NotImplementedException is perfectly valid in these cases.

Assignment9/WpfApp/WpfApp/ContactViewModel.cs
```diff
@@ -0,0 +1,42 @@
...
+        public string? LastName { get; set; }
+        public string? PhoneNumber { get; set; }
+        public string? EmailAddress { get; set; }
+        public string? TwitterName { get; set; }
+
+        private string _LastModifiedTime = DateTime.Now.ToString();
+        public string LastModifiedTime
+        {
+            get => _LastModifiedTime;
+            set => _LastModifiedTime = DateTime.Now.ToString();
```
Keboo, line 22: This setter is odd since it does use the incoming `value`. I suspect this is the reason for your issue when editing a new contact.

Assignment9/WpfApp/WpfApp/MainWindowViewModel.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        private string _EditSaveText = "Edit";
+        public string EditSaveText
+        {
+            get => _EditSaveText;
+            set => SetProperty(ref _EditSaveText, value);
+        }
+
+        private void UpdateListBox()
```
Keboo, line 63: nit: In general I try to keep UI elements out of my view models (even by name)
```suggestion
        private void UpdateContactsList()
```

Assignment9/WpfApp/WpfApp/GenericCommand.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Windows.Input;
+
+namespace WpfApp
+{
+    public class GenericCommand : ICommand
+    {
+        public Action Action { get; }
```
Keboo, line 8: This could have been private as an implementation detail. To test you would just invoke the Execute method directly.

Keboo (commented):
# Assignment
- Can create a new contact: ✔
- Can delete a contact: ✔

## Extra Credit

PR #84, opened 2021-03-19
Assignment9

Sithale:
Program should function, and tests were made. Still missing tests for ContactViewModel.
No extra credit was attempted.

### DIFF
--- a/Assignment9/Assignment9.Tests/MainWindowViewModelTests.cs
+++ b/Assignment9/Assignment9.Tests/MainWindowViewModelTests.cs
@@ -1,0 +1,79 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Linq;
+
+namespace Assignment9.Tests
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+        [TestMethod]
+        public void MainWindowViewModel_UsingCreateAndDeleteCommands_ContactIsCreatedAndDeleted()
+        {
+            // Arrange
+            MainWindowViewModel model = new();
+
+            // Act
+            int before = model.Contacts.Count;
+            var contact = model.Contacts.Last();
+
+            model.SelectedContact = contact;
+            model.DeleteContactCommand.Execute(null);
+
+            // Assert
+            Assert.AreEqual<int>(before - 1, model.Contacts.Count);
+            Assert.IsFalse(model.Contacts.Contains(contact));
+
+        }
+
+        [TestMethod]
+        public void MainWindowViewModel_UsingEditContactCommand_IsBeingEditedReturnsTrue()
+        {
+            // Arrange
+            MainWindowViewModel model = new();
+
+            // Act
+            model.IsBeingEdited = false;
+            model.EditContactCommand.Execute(null);
+
+            // Assert
+            Assert.IsTrue(model.IsBeingEdited);
+
+        }
+
+        [TestMethod]
+        public void MainWindowViewModel_UsingSaveContactCommandToExitEditing_IsBeingEditedReturnsFalse()
+        {
+            // Arrange
+            MainWindowViewModel model = new();
+
+            // Act
+            model.IsBeingEdited = true;
+            model.SelectedContact = model.Contacts.First();
+            model.SaveContactCommand.Execute(null);
+
+            // Assert
+            Assert.IsFalse(model.IsBeingEdited);
+
+        }
+
+        [TestMethod]
+        public void MainWindowViewModel_UsingSaveContactCommand_LastModifiedIsUpdated()
+        {
+            // Arrange
+            MainWindowViewModel model = new();
+
+            // Act
+            model.SelectedContact = model.Contacts.First();
+            DateTime originalDate = DateTime.Now;
+            model.SelectedContact = model.Contacts.First();
+            model.SelectedContact.LastModified = originalDate;
+            model.SaveContactCommand.Execute(null);
+
+            // Assert
+            Assert.AreNotEqual<DateTime>(originalDate, (DateTime)model.SelectedContact.LastModified);
+
+        }
+    }
+}
+
--- a/Assignment9/Assignment9/ContactViewModel.cs
+++ b/Assignment9/Assignment9/ContactViewModel.cs
@@ -1,0 +1,71 @@
+﻿using System;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+
+namespace Assignment9
+{
+    public class ContactViewModel : INotifyPropertyChanged
+    {
+        public event PropertyChangedEventHandler PropertyChanged;
+        private string? _FirstName;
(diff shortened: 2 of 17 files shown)


Assignment9/Assignment9/ContactViewModel.cs
```diff
@@ -0,0 +1,71 @@
...
+        public string? EmailAddress
+        {
+            get => _EmailAddress;
+            set => SetProperty(ref _EmailAddress!, value!);
+        }
+
+        public string? TwitterName
+        {
+            get => _TwitterName;
+            set => SetProperty(ref _TwitterName!, value!);
```
Keboo, line 44: It would have been better to make the parameters of `SetProperty` nullable rather than using the null forgiveness operator on all of the invocations.

Assignment9/Assignment9/ContactViewModel.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            get => _LastModified;
+            set
+            {
+                _LastModified = value;
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(LastModified)));
+            }
+        }
+
+        private void SetProperty(ref string field, string newValue, [CallerMemberName] string propertyName = "")
```
Keboo, line 57: This is not quite the correct implementation. The correct implementation is like the one you have in MainWindowViewModel.

Assignment9/Assignment9/ContactViewModel.cs
```diff
@@ -0,0 +1,71 @@
...
+            set => SetProperty(ref _TwitterName!, value!);
+        }
+
+        public DateTime LastModified
+        {
+            get => _LastModified;
+            set
+            {
+                _LastModified = value;
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(LastModified)));
```
Keboo, line 53: You actually only want to raise the `PropertyChanged` if the value has actually changed.

Assignment9/Assignment9/EmptyStringToVisibilityConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (string.IsNullOrEmpty((string)value))
+                return Visibility.Collapsed;
+
+            else
+                return Visibility.Visible;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
```
Keboo, line 19: This is not quite right. The value passed into `ConvertBack` will be the return value from the `Convert` method. In this case there is not really a rational invert of the `Convert` method. For these one-way converters it is perfectly fine to simply leave the `ConvertBack` method throwing NotImplementedException.

Assignment9/Assignment9/ContactViewModel.cs
```diff
@@ -0,0 +1,71 @@
...
+            set => SetProperty(ref _EmailAddress!, value!);
+        }
+
+        public string? TwitterName
+        {
+            get => _TwitterName;
+            set => SetProperty(ref _TwitterName!, value!);
+        }
+
+        public DateTime LastModified
```
Keboo, line 47: It doesn't look like this is bound in the UI anywhere.

Assignment9/Assignment9/MainWindowViewModel.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        public bool IsBeingEdited
+        {
+            get => _IsBeingEdited;
+            set => SetProperty(ref _IsBeingEdited, value);
+        }
+
+        private void NewContact()
+        {
+            ContactViewModel NewContact = new ContactViewModel() { FirstName = "Temp", LastName = "Name", LastModified = DateTime.Now };
```
Keboo, line 77:
```suggestion
            ContactViewModel newContact = new ContactViewModel() { FirstName = "Temp", LastName = "Name", LastModified = DateTime.Now };
```

Assignment9/Assignment9/MainWindowViewModel.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        private bool _IsBeingEdited;
+        private ContactViewModel _SelectedContact;
+
+        public MainWindowViewModel()
+        {
+            NewContactCommand = new RelayCommand(NewContact, () => true);
+            EditContactCommand = new RelayCommand(EditContact, () => true);
+            SaveContactCommand = new RelayCommand(SaveContact, () => true);
+            DeleteContactCommand = new RelayCommand(DeleteContact, CanDeleteContact);
```
Keboo, line 29: If you delete all of the contacts you will notice that your DeleteContactCommand doesn't actually disable the button. The reason is because you don't raise the CanExecuteChanged event on the DeleteContactCommand. Commands that have CanExecute delegate need to take cause to appropriately raise that event.

Assignment9/Assignment9/RelayCommand.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Windows.Input;
+
+namespace Assignment9
+{
+    public class RelayCommand : ICommand
+    {
+        public event EventHandler CanExecuteChanged;
+        public void RaiseCanExecuteChanged() => CanExecuteChanged?.Invoke(this, EventArgs.Empty);
+        public bool CanExecute(object parameter) => CanExecuteDelegate.Invoke();
+        public void Execute(object parameter) => ExecuteDelegate.Invoke();
```
Keboo, line 11: There is a warning generated here because the parameter types don't match the nullability of the interface.
```suggestion
        public bool CanExecute(object? parameter) => CanExecuteDelegate.Invoke();
        public void Execute(object? parameter) => ExecuteDelegate.Invoke();
```

Assignment9/Assignment9.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,79 @@
...
+using System;
+using System.Linq;
+
+namespace Assignment9.Tests
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+        [TestMethod]
+        public void MainWindowViewModel_UsingCreateAndDeleteCommands_ContactIsCreatedAndDeleted()
```
Keboo, line 11: There is something off with the name of this test. It looks like the `NewContactCommand` is not actually used here (and it appears that it is not tested). In general you have to avoid needing conjunctions (such as And, Or, etc.) in your test names.

Keboo (commented):
# Assignment
- All view models are tested: ✔/❌ - NewContact and CanDeleteContact are not tested. 
- Can create a new contact: ✔
- Can delete a contact: ✔ - Yes though the delete command does not properly raise its CanExecuteChanged event when the last contact is deleted.

## Extra Credit

Keboo:
@Sithale I am not seeing a code review from you for assignment 9. Am I missing it?

PR #85, opened 2021-03-19
Assignment9

alexvthornton:
partner: @kaykay38


Assignment9/Assignment9.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,143 @@
...
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+        [TestMethod]
+        public void Contacts_ReturnsObservableContactCollection()
+        {
+            var viewModel = new MainWindowViewModel();
+
+            Assert.AreEqual(typeof(ObservableCollection<Contact>), viewModel.Contacts.GetType());
```
Keboo, line 17: A better way to do type checking is by simply using the `is` keyword. Checks with `.GetType()` quickly get problematic when you start dealing with derived types.
```suggestion
            Assert.IsTrue(viewModel.Contacts is ObservableCollection<Contact>);
```

Assignment9/Assignment9.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            viewModel.Edit = true;
+
+            Assert.IsTrue(viewModel.Edit);
+        } 
+        [TestMethod]
+        public void EditCommand_ReturnsRelayCommand()
+        {
+            var viewModel = new MainWindowViewModel();
+            Assert.AreEqual(typeof(RelayCommand), viewModel.EditCommand.GetType());
+        }
```
Keboo, line 59: In general you don't need tests like this (the same applies to `EditCommand_NotNull`). Typically the "success" case like this gets implicitly covered when you write a test that uses the command. So these tests can be omitted. For example your `EditCommand_True_False` also validates that the EditCommand is properly initialized.

Assignment9/Assignment9/MainWindowViewModel.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    public class MainWindowViewModel : INotifyPropertyChanged
+    {
+        public ICommand EditCommand { get; }
+        public ICommand NewContactCommand { get; }
+        public ICommand DeleteContactCommand { get; }
+        public event PropertyChangedEventHandler PropertyChanged;
+        public ObservableCollection<Contact> Contacts { get; } = new ();
+
+        private bool _edit = false;
```
Keboo, line 19: Don't need to initialize to default value.
```suggestion
        private bool _edit; 
```

Keboo (commented):
# Assignment
- WPF Project created with nullability, and analyzers enabled: ❌ - Nullability was not enabled.
- Can create a new contact: ✔
- Can delete a contact: ✔

## Extra Credit

PR #86, opened 2021-03-19
Assignment9

jschofield1:
Partnered with @coproduct


jschofield1:
Assignment
•	WPF Project created with nullability, and analyzers enabled: ✔
•	App uses MVVM pattern (MainWindowViewModel): ✔
•	All view models are tested: ✔
•	App shows a list of contacts: ✔
•	Can create a new contact: ✔
•	Can delete a contact: ✔
•	Can view a contact's details: ✔
•	Can edit a contact's details: ✔
•	Contact fields that are not specified are not shown while viewing using triggers: ✔
•	Contact contains all required fields: ✔
Extra Credit
•	In addition to the default contact fields, allow for adding an arbitrary set of additional fields: ❌ 
•	App allows for savings and loading contacts using Entity Framework Core (simply use SQLite): ❌

Assignment9/Assignment9/Assignment9.Tests/ContactTests.cs
```diff
@@ -0,0 +1,64 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace Assignment9.Tests
+{
+    [TestClass]
+    public class ContactTests
```
Keboo, line 7: These are fine, however most of these tests are simply validating that auto properties work. In general you only need to test the code you write. You can consider auto property work as designed and as such don't need to be directly tested. A more interesting test would have been validating that setting a new value `LastModifiedTime` raised the `PropertyChanged` event.

Assignment9/Assignment9/Assignment9.Tests/CommandTests.cs
```diff
@@ -0,0 +1,31 @@
...
+        public void CanExecuteDelegate_CanExecuteReturnsTrueWhenAbleToExecute()
+        {
+            Command relay = new(() => System.Console.WriteLine("Testing"), () => true);
+            Assert.IsTrue(relay.CanExecute(null));
+        }
+
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void RelayCommand_PassInNullValues_ThrowNullArgumentException()
```
Keboo, line 26: This technically only tests the null case for the `executeDelegate` since it will never get to checking the `canExecuteDelegate` since the first one will throw first. Typically you would write a second test to check the null cases for the `canExecuteDelegate`.

Keboo (commented):
# Assignment
- App uses MVVM pattern (MainWindowViewModel): ✔/❌ - Contact has the PropertyChanged event but does not implement INotifyPropertyChanged.
- Can create a new contact: ✔
- Can delete a contact: ✔

## Extra Credit

PR #87, opened 2021-06-28
Jenny assn5

jmeadowcroft:


Keboo:
@jmeadowcroft it looks like your branch also includes all of your Assignment 4 stuff? Do you mind removing those files so it is easier to review?

jmeadowcroft:
Done

Assignment5/GenericsHomwork.Tests/GenericsHomwork.Tests.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
Keboo, line 3: You want to have the same nullability property set here.
`<Nullable>enable</Nullable>`

Assignment5/GenericsHomwork.Tests/NodeTests.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace GenericsHomwork.Tests
+{
+    [TestClass]
+    public class NodeTests
+    {
+        [TestMethod]
+        public void Constructor_ValidInput_Success()
+        {
+            int number = 4;
+            Node<int> testNode = new Node<int>(number);
```
Keboo, line 14: nit: You can leverage C# [target typed new expressions](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/target-typed-new) to simplify this.
```suggestion
            Node<int> testNode = new(number);
```

Assignment5/GenericsHomwork.Tests/NodeTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            Assert.IsNotNull(testNode);
+        }
+        [ExpectedException(typeof(ArgumentNullException))]
+        [TestMethod]
+        public void Constructor_NullInput_Success()
+        {
+            int? number = null;
+            Node<int?> testNode = new Node<int?>(number);
```
Keboo, line 22: Because `testNode` is not used you can use a discard. In C# 9, this actually will tell the compiler to not even both to generate a variable for this (previously versions of C# this would still be a variable with the name of "_").
```suggestion
            _ = new Node<int?>(number);
```

Assignment5/GenericsHomwork.Tests/NodeTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        [TestMethod]
+        public void Constructor_NullNext_Success()
+        {
+            int number = 4;
+            Node<int> testNode = new Node<int>(number, null);
+            Assert.IsNotNull(testNode);
```
Keboo, line 29: In this test case I would expect you to test against `testNode.Next` rather than `testNode` itself. Testing that `testNode` is not null really just tests that constructors in C# work.

Assignment5/GenericsHomwork.Tests/NodeTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        [TestMethod]
+        public void Insert_ValidInput_Success()
+        {
+            int number = 4;
+            Node<int> testNode = new Node<int>(number);
+            testNode.Insert(number);
+            Assert.IsNotNull(testNode);
```
Keboo, line 37: I would expect this test to also tests against the `testProperty.Next`. I would also insert a different value than what was passed in the constructor; just to make it easier to know which node we are looking at.

Something more like (assumes Value property getter is made public):
```suggestion
            Node<int> testNode = new(4);
            testNode.Insert(42);
            Assert.AreEqual(4, testNode.Value);
            Assert.AreEqual(42, testNode.Next.Value);
```

Assignment5/GenericsHomwork.Tests/NodeTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+        [ExpectedException(typeof(ArgumentNullException))]
+        [TestMethod]
+        public void Insert_NullInput_Success()
+        {
+            int? number = 4;
+            Node<int?> testNode = new Node<int?>(number);
+            number = null;
+            testNode.Insert(number);
```
Keboo, line 46: nit: My preference is to not reuse variables for different purposes. In this case since you really just have a single usage of each (constructor value and insert call), and the variable name is not adding much value (clarity of what the value means) you could just inline it.

NB: I also added the [null-forgiving operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving#:~:text=Available%20in%20C%23%208.0%20and%20later%2C%20the%20unary,prefix%20%21%20operator%20is%20the%20logical%20negation%20operator.) which would be needed to handle the warning that comes when you enable nullability.
```suggestion
            Node<int?> testNode = new Node<int?>(4);
            testNode.Insert(null!);
```

Assignment5/GenericsHomwork.Tests/NodeTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            Assert.AreEqual(expectedOutput, output);
+        }
+        [TestMethod]
+        public void DestroyLinks_ValidNode_Success()
+        {
+            Node<int> testNode = new Node<int>(4);
+            Node<int> testNodeTwo = new Node<int>(5,testNode);
+            Node<int> testNodeThree = new Node<int>(6,testNodeTwo);
+            Node<int> testNodeFour = new Node<int>(7,testNodeThree);
```
Keboo, line 64: `testNodeFour` appears to be unused.
```suggestion
            _ = new Node<int>(7,testNodeThree);
```

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        public Node(T value)
+        {
+            if (value is null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+            Value = value;
+            Next = this;
+        }
```
Keboo, line 41: You can chain these constructors together to avoid needing to duplicate logic. Like this:
```suggestion
        public Node(T value)
            : this(value, null)
        { }
```

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        public Node<T> Next { get; private set; }
+        T Value { get; set; }
+
+        public Node(T value, Node<T> next)
+        {
+            if (value is null)
```
Keboo, line 13: Technically this is not needed as the assignment calls out that you don't need any validation on the value.

> Include a constuctor that takes a value.  (No validation is necessary on the value). ❌✔

global.json
```diff
@@ -3,5 +3,8 @@
     "vs": {
       "version": "16.8"
     }
+  },
+  "sdk": {
+    "version": "5.0.200"
```
Keboo, line 8: Any reason for pinning this to an exact version?

Assignment5/GenericsHomework/Node.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace GenericsHomework
+{
+    public class Node<T>
+    {
+        public Node<T> Next { get; private set; }
```
Keboo, line 8: Technically the assignment asks for the validation of this property to be in the setter. It appears that you put this logic as part of the constructor. Though this works, it also means that in theory if someone were to work with the code later, they might accidently assign a null to the property. By putting it in the setter it consolidates the validation logic into one place.

> In addition to non-null validation, the body of the `Next` property should insert the next `Node` (the `value`) into the list. ❌✔

Keboo (commented):
- Create a *class library* project called "GenericsHomework.". ✔
- Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item.
  - Define the `Node` class ✔
  - Include a constuctor that takes a value.  (No validation is necessary on the value). ✔
  - Add a `ToString()` override that writes out the value's `ToString()` result. ✔
  - Add a `Next` property that references the next node or else refers back to itself if there are no other nodes in the list. ✔
    - The `Next` property should be non-nullable (careful to follow the non-nullable property guidelines) ✔
    - The `Next` property setter should be private. ✔
    - In addition to non-null validation, the body of the `Next` property should insert the next `Node` (the `value`) into the list. ❌
  - Add an `Insert` method that takes a value and inserts a new `Node` instance after the current node (by invoking the `Next` property).✔
  - Add a clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be
  concerned with the following:
    - Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and therefore may never be garbage collected. ✔

